PR #51, opened 2021-11-10
Implement streaming annotations

### LINKED ISSUE #40: Switch to streaming annotation for WaveformModel annotate
The current implementation of annotate (i) first preprocesses the full stream before the next steps and (ii) treats each segment of a stream separately, i.e., each contiguous segment of waveforms from one instrument. (i) leads to higher memory consumption, as copies of the full stream are required. This will be an issue for large inputs. (ii) decreases efficiency of the prediction for streams with many small segments as batching is not used efficiently in the model calculation. Both issues could be solved by a streaming annotation.

yetinam:
This commit implements streaming processing for WaveformModels. This has several advantages:

- Enables parallelization between and within each step
- Allows batching across multiple stations, thereby improving GPU utilization
- Reduces memory consumption for point prediction models in case of long traces

As this only modifies the backend, but does not introduce any changes in the user interface, I'd suggest to publish it as 0.1.8 (instead of 0.2.0), even though it's technically not a bugfix. What do you think @jawooll ? In addition, because it's a somewhat more complicated change, I'd ask you to have a look at it.

Closes #40

### DIFF
--- a/.github/workflows/main_push.yml
+++ b/.github/workflows/main_push.yml
@@ -27,7 +27,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
-        python-version: ['3.7', '3.8', '3.9', '3.10']
+        python-version: ['3.7', '3.8', '3.9']
         
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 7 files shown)


jawooll:
Hey thanks for implementing this! I agree with publishing to 0.1.8 as the user interface hasn't changed - makes sense. I'll take a look and get back to you.

jawooll:
I've run across datasets and waveforms (similar to the unit test cases you implement but just for seismic examples with various models). All looks good. Merging now.

PR #55, opened 2021-12-08
Additional example workflows

jawooll:
Following feedback from an internal workshop, this PR just adds a simple workflow to the examples where instead of running through the components in-depth, an applied example demonstrates how to use SeisBench to apply ML models to pick on seismic streams.


seisbench/generate/augmentation.py
```diff
@@ -101,6 +101,28 @@ def __str__(self):
...
+    :type key: str, tuple[str, str]
+
+    """
+
+    def __init__(self, key=("X", "Xc")):
+        self.key = key
+
+    def __call__(self, state_dict):
+        x, metadata = state_dict[self.key[0]]
+        state_dict[self.key[1]] = (x, metadata)
```
yetinam, line 120: I would suggest to make deep instead of shallow copies here, i.e., change the lower line to `... = (x.copy(), copy.deepcopy(metadata))`.

tests/test_generate.py
```diff
@@ -1,3 +1,4 @@
+from seisbench import data
```
yetinam, line 1: Looks like an unused import. Could you check and delete if it's indeed unused?

yetinam:
Hey Jack,
thanks for the update! It looks good to me, I just have a few minor points listed below (and two comments in the code). Feel free to merge the PR once you addressed them.

- I really like the "Open in colab" badge on top of the notebook. Could you add a similar badge to all other notebooks as well? Also make sure to change the link to the main branch before merging.
- Please add the example to the list of examples in both the readme and the documentation.
- For making it look a bit prettier I'd rerun the notebook and push a version without the installation output in the first cell.
- I'd suggest sticking to the naming scheme with numbers in the beginning. I'm not quite sure which number we should go for, so feel free to pick any ;-)
- As we already have tree models available, I think it would be nice to show the predictions in comparison. This should only require minor tweaking to the plotting function. Maybe we could make one waveform panel and three prediction panels with height rations (3, 1, 1, 1) or something similar?

Last note, I think we should merge into `maintenance_0.1.x` instead of `main` and then merge into `main` from there. The new example does not really need to be in the maintenance, but I think it would be great having the copy augmentation there, so that it will be part of release 0.1.10 eventually. Otherwise you could split the PR in two, one for the copy augmentation and one for the example and merge only the copy augmentation to maintenance, but from my side I'd say it's not worth the hassle.

yetinam (commented):
See comment before

seisbench/generate/augmentation.py
jawooll: Good point, done and added test to check it is working as expected.

tests/test_generate.py
jawooll: Well spotted, thanks! Removed.

jawooll:
Ok great thanks for the quick review, to just go through your points: 

- [X] Open in Colab - added this to all examples, the link will work when merged into main branch
- [X] Updated README and documentation (as above with link working when on main branch)
- [X] Removed cell output
- [X] I've gone for 01d, but unsure how this serial way of naming examples will fit in we expand the example workflows. Maybe we could have 01 for simple workflows, 02 for intermediate and 03 for advanced?
- [ ] (Change fig. in notebook) I can add this in now

Ok and for the final point, let's reject this PR and submit a new one to the maintenance branch.

yetinam:
I like your numbering idea for the examples. I created an issue (#56) for this. We can take care of it once this PR is merged.

Regarding the target branch, I switched it to `maintenance_0.1.x` now, so no need for creating a new PR.

yetinam:
Turns out I was very ambiguous with my point on comparing predictions. Sorry for that. I meant comparing across models, not the different prediction curves of one model. I implemented and pushed the change now.

jawooll:
All looks good! I’ve gone ahead and merged with all edits now integrated - thanks

Issue #69, opened 2022-03-09
Any examples for saving a model after training so it can later be loaded with from_pretrained() or similar?


alomax:
I have been modifying `03a_training_phasenet.ipynb` (e.g. with different augmentations) and then (re)training a provided SeisBench pretrained model. But I cannot see how to save the newly trained model for later use, perhaps simply by calling `from_pretrained()`. 

I tried the simple use of:
`torch.save(model.state_dict(), PATH)`
without attempting to create the `[name].json` config file (I do not find any SeisBench code in the repository that writes the json config file.)
But this does not seem to be sufficient for later use of the trained model:

```
model_copy = sbm.PhaseNet()
model_copy.load_state_dict(torch.load(PATH))
preds = model_copy.annotate(stream)
```
seems to work, but plotting the results gives an error:

```
---------------------------------------------------------------------------
KeyError                                  Traceback (most recent call last)
Input In [41], in <module>
     15     if pred_class == "N":
     16         # Skip noise traces
     17         continue
---> 18     c = color_dict[pred_class]
     19     ax[i + 1].plot(offset + pred_trace.times(), pred_trace.data, label=pred_class, c=c)
     20 ax[i + 1].set_ylabel(model)

KeyError: '0'

```
Perhaps the missing `[name].json` config file is the culprit.

Is there any code or examples that show how to save a newly trained model so it can be easily used later in SeisBench in a manner similar to `from_pretrained()` for provided SeisBench pretrained models?

Thanks,
Anthony

jawooll:
Hi Anthony,

Thanks for raising this issue, I can't seem to re-create the error as you describe above. The following code should be self-contained and load in the pre-trained model weights as you describe after saving them locally. Maybe it was just the filetype in the path which need appropriately specifying (.pt):

```python
import torch
import seisbench.models as sbm
from obspy.clients.fdsn import Client
from obspy import UTCDateTime

# Load in pre-trained PhaseNet model
pn_model = sbm.PhaseNet.from_pretrained("stead")
# Let's save the weights locally 
torch.save(pn_model.state_dict(), './abc.pt')

# Copy the model architecture
model_copy = sbm.PhaseNet()
# Load the now locally stored weights in
model_copy.load_state_dict(torch.load('./abc.pt'))

# Get a stream
client = Client("INGV")
t = UTCDateTime(2009, 4, 6, 1, 30)
[...]

PR #71, opened 2022-03-16
Add ability to save and load model locally (#69)

jawooll:
- Introduces functionality for saving and loading models from local paths
- Saved model formats adhere to the currently used structure (.json and .pt files)
- Config parameters are obtained during constructor call of ```WaveformModel```

With the current implementation, drawbacks include the fact that if parameters are overwritten with new variables after construction they will not be saved, and the fact that non-serializable object arguments (e.g. functions) cannot be wrote to json. 

I'd be happy to work further on how we can tackle them final two points, but it was worth having a discussion of the current approach first. Let me know what you think!


seisbench/models/base.py
```diff
@@ -192,6 +193,113 @@ def download_metadata_callback(metadata_path):
...
+
+        model.load_state_dict(model_weights)
+
+        return model
+
+    def save(self, path=None, name=None):
+        """
+        Save a SeisBench model locally.
+
+        SeisBench models are stored inside the directory 'path'. SeisBench models are saved in 2 parts,
```
yetinam, line 256: So far, we do not identify models by their path but by their name. For example, the pretrained models for GPD are all in the same path, but called `stead.json/stead.pt`, `ethz.json/ethz.pt`, ... . I'd suggest to keep it that way. This differs from datasets, where we assumed one directory per dataset, but I think this is reasonable because model weights are a lot smaller in size.

For saving/loading this would mean that a user just provides one path (to either json, pt, or simply without extension) and the files are than loaded automatically.

seisbench/models/base.py
```diff
@@ -192,6 +193,113 @@ def download_metadata_callback(metadata_path):
...
+        model = cls(**cls_args)
+
+        model._weights_metadata = weights_metadata
+        model._parse_metadata()
+
+        model.load_state_dict(model_weights)
+
+        return model
+
+    def save(self, path=None, name=None):
```
yetinam, line 252: I'd make the `path` argument (see also comment below) mandatory. I think it forces users to best practice if they have to provide the path explicitly.

seisbench/models/base.py
```diff
@@ -192,6 +193,113 @@ def download_metadata_callback(metadata_path):
...
+        The model config should contain the following information, which is automatically created from
+        the model instance state:
+            - "docstring": A string documenting the pipeline. Usually also contains information on the author.
+            - "model_args": Argument dictionary passed to the init function of the pipeline.
+            - "seisbench_requirement": The minimal version of SeisBench required to use the weights file.
+            - "default_args": Default args for the :py:func:`annotate`/:py:func:`classify` functions.
+
+        Non-serlisable arguments (e.g. functions) cannot be saved to JSON, so are not converted.
+
+        :param path: Define the output model directory. If none is provided, the path will create a directory
```
yetinam, line 269: I'd not create the directory if missing but throw an exception, just as the python `write` function.

seisbench/models/base.py
```diff
@@ -192,6 +193,113 @@ def download_metadata_callback(metadata_path):
...
+            path = Path(f"{name}")
+        path.mkdir(parents=True, exist_ok=True)
+
+        model_args = {
+            k: v
+            for k, v in self._argspec.items()
+            if k not in ("__class__", "self", "default_args") and not callable(v)
+        }
+        model_metadata = {
+            "docstring": self.__class__.__doc__,
```
yetinam, line 290: I think using `__class__.__doc__` is not the optimal solution. I'd rather add the docstring as a parameter to the save function. Maybe it could be empty by default but seisbench returns a warning if you actually keep it empty?

seisbench/models/base.py
```diff
@@ -192,6 +193,113 @@ def download_metadata_callback(metadata_path):
...
+            name = self.name.lower()
+
+        if not path:
+            path = Path(f"{name}")
+        path.mkdir(parents=True, exist_ok=True)
+
+        model_args = {
+            k: v
+            for k, v in self._argspec.items()
+            if k not in ("__class__", "self", "default_args") and not callable(v)
```
yetinam, line 287: I think there should be a test for this. In addition, could we raise warnings if something is discarded here?

seisbench/models/base.py
```diff
@@ -192,6 +193,113 @@ def download_metadata_callback(metadata_path):
...
+            "seisbench-requirements": seisbench.__version__,
+            "default_args": self._argspec["default_args"],
+        }
+        # Save weights
+        torch.save(self.state_dict(), path / f"{name}.pt")
+        # Save model metadata
+        with open(path / f"{name}.json", "w") as json_fp:
+            json.dump(model_metadata, json_fp)
+
+        seisbench.logger.info(f"Saved {name} model at {path.absolute}")
```
yetinam, line 301: I think this is not really necessary. Maybe we could remove it or move it to debug level? Even though, I admit, our usage of `INFO` and `DEBUG` so far is not really consistent.

seisbench/models/base.py
```diff
@@ -192,6 +193,113 @@ def download_metadata_callback(metadata_path):
...
+        For more information on the SeisBench model format see py:func:`save`.
+
+        :param path: Define the path to the SeisBench model directory.
+        :type path: pathlib.Path
+        :return: Model instance
+        :rtype: SeisBenchModel
+        """
+        model_files = os.listdir(path)
+
+        if sum([file.endswith("pt") for file in model_files]) > 1:
```
yetinam, line 210: There should be a check if their is actually a weights file. The json file (so far) is not mandatory.

seisbench/models/base.py
```diff
@@ -192,6 +193,113 @@ def download_metadata_callback(metadata_path):
...
+        def _get_args(obj):
+            signature = inspect.signature(obj)
+            args = {k: v._default for k, v in signature.parameters.items()}
+            if "kwargs" in args.keys():
+                del args["kwargs"]
+            return args
+
+        model_args = weights_metadata.get("model_args", {})
+
+        # Overwrite default class args from config model args
```
yetinam, line 239: I'm not really sure why this part is needed. Could you explain this?

yetinam (changes requested):
Hey @jawooll,
thanks a lot for the commits. I've now gone through the code. I added a few principled comments which are obviously subject to discussion. I have two more general points:

- I think this PR should also include a change to `from_pretrained` so that it uses the new `load` method. Similarly, the documentation should be deduplicated.
- We should updated the documentation [here](https://seisbench.readthedocs.io/en/stable/pages/models.html#loading-pretrained-models) to incorporate the possibility of local saving and loading.

seisbench/models/base.py
jawooll: Thanks for the comment. This makes sense and it would then just be a case of aligning both load / save functions to use ```path``` param. I can edit the load fn to do this.

seisbench/models/base.py
jawooll: Sure this is a good point! - I can add the logic for both.

seisbench/models/base.py
jawooll: Just on this, what's the reasoning behind dropping the docstring? My thinking was that as we are always saving / loading the config to reproduce some child '_model_' class, inheriting from SeisBench model, the docstring of this child class will stay the same? Is there a case I'm missing where we would want a different docstring?

yetinam: The `docstring` argument should describe the weights, not the model. The model is already described in the class/documentation. Saving a property of the class in the json file would be redundant as the json can anyhow only be loaded if you have the implementation of the class available.

seisbench/models/base.py
jawooll: Ok, so this comes about because when we save out the model we do this by tracking all the state during the ```WaveformModel``` constructor call with ```self._argspec = locals()```. 

This means there are extra configuration parameters in the ```self._argspec``` dict which belong to the parent class ```WaveformModel```. We do this here as we need to provide a layer of abstraction so that all inherited models do not have to implement the ```locals()``` call themselves. 

We then just 'cherry-pick' or extract out the arguments needed to pass to the model constructor only.

seisbench/models/base.py
jawooll: Ah ok got you! Yeah, I was thinking this was a bit of circular argument when writing, but that was all coming from my misunderstanding of 'docstring' - I can update this now, cheers!

seisbench/models/base.py
jawooll: After reading your solution to this problem of identifying what arguments are relevant for saving below, the above code is just my current approach to do this. With the outlined drawback of not being able to detect parameters changed after model construction a current limitation during saving.

seisbench/models/base.py
```diff
@@ -339,6 +447,7 @@ def __init__(
         self._annotate_functions = self._annotate_function_mapping.get(
             output_type, None
         )
+        self._argspec = locals()
```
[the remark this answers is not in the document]
jawooll: I would be slightly apprehensive about making the users overwrite an abstract method to save any new model. I think it it just another step which adds complexity to our model API. I may have a fix for this to do it automatically, it is through object introspection, and slightly verbose but it does manage to avoid the issue of users having to perform extra steps in order to save a model.

tests/test_models.py
```diff
@@ -1047,3 +1048,81 @@ def test_eqtransformer_annotate_window_post():
     assert np.isnan(blinded[:100]).all()
     assert (blinded[100:900] == 1).all()
     assert np.isnan(blinded[900:]).all()
+
+
+def test_model_local_saving(tmp_path):
```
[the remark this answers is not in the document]
jawooll: Good point I will add these in. I've gone about trying to account for parameter overwritten with a slightly tweaked approach, so if your happy with this method, I'll continue with the unit tests, but it would be worth checking the logic first.

jawooll:
Thanks for the review. I've gone ahead and implementing many of the minor logic changes and edited the unit tests. One major change here that I first want to run by you before continuing is that **I have tackled the problem of accounting for parameters being overwritten after object construction** which we could not do through the use of ```locals()```. It works as follows:

1. During model saving, get the input arguments to the model (_model_args_).
2. Also obtain the entire state of the object via ```self.__dict__```.
3. Check whether any input parameters have been updated in state via ```_update_object_args()```

For 3. in order to map the input parameters to instance attributes, I assume they have the same name (e.g. self.a = a or self._a = a). If they cannot be detected, a warning is logged and it tries to revert back to default values. If this parameter is a required parameter, without a default value, the ```save()``` method raises an error. 

As mentioned this allows us to account for the latest changes to some model, which may of been changes after initialisation. It also makes sure that the users don't need to add any more steps on their end. Drawbacks are the assumption that the instance variable is the same name as the input variable, but I think this is good-practice, and users will encounter appropriate errors or warnings if this happens. Let me know is this seems feasible.

yetinam:
Hey Jack,

thanks for continuing on the PR. I started looking at it now. While I agree that it would be nice not to put the work for serialization on the person implementing the model, the current approach has an issue with inheritance. Here is a small code snippet visualizing the issue:

```python
model = GPD(component_order="NZE")
model.save(".")
model_loaded = model.load(".")
model_loaded.component_order  # ZNE
```

The current implementation can only save parameters that are part of the outermost constructor, but does ignore the constructor args passed to `WaveformModel` through the `kwargs`. I remembered that pytorch lightning has a very similar function for saving constructor args called `save_hyperparameters` that takes into account the inheritance. You can find it here: https://github.com/PyTorchLightning/pytorch-lightning/blob/4277845fa7dc3831660143a90aa3757528a588c5/pytorch_lightning/utilities/parsing.py#L204

Regarding the issue about parameters changed after construction, I generally like the idea to detect this automatically. However, I'm not quite sure whether there are maybe good reasons to deviate from this pattern sometimes and I think it kind of adds a somewhat hidden assumption on the model implementations. I think requiring users to implement a function for serialization could be the clearer option.

I have another somewhat central thing, which is about the path names. I think we had a bit of a misunderstanding here. I was suggesting to identify each model by a path including a name for the weight. Example code would be:
```python
model.save("path/to/my_model_weights", docstring="Model with my weights)  # Writes files "path/to/my_model_weights.pt" and "path/to/my_model_weights.json"
```
This allows users to store multiple models in the same directory. This is necessary because we already use this naming convention for the pretrained weights and I think we should not break that. You can have a look at you local SeisBench cache.

Given that these points will require some discussion and probably larger changes to the PR, I'll wait with the detailed review until we have agreed on a strategy for the points here.

jawooll:
Thanks for the quick feedback! All points make sense, just a bit of explanation from my side about how this inheritance problem manifested; attempting to pass all parameters through to parent constructors as kwargs raises an error 

```python
model = cls(**model_args)

>>> "TypeError: __init__() got multiple values for keyword argument 'output_type'"
```
So the current approach only allows for changing variables which belong to the child model and not ```WaveformModel``` or any other parent class. As I have not yet found a way around this, the parent class parameters are 'hardcoded' currently. 

I've just had a look at the PyTorch Lightning approach, and this seems very very similar (if not identical) to what we are trying to do, so I'll take a look how they went about this (thanks for the pointer).

yetinam:
I see the issue. It's hard to identify which parameters of the super constructor are actually user-provided and which are hard-coded... Guess we'll need to figure this out. Maybe the pytorch lightning thing can help there as well.

jawooll:
@yetinam I just want to touch base here before continuing. With your outlined implementation for doing local-model-saving-and-loading on the users' side via an overwritable abstract method. Is this how you envisage it? 

It is still quite a difficult problem to generalise as there are a number of checks that have to be performed before writing different parameters to the JSON file. Because of these strict controls on what metadata information from a model state you can write out in the .json metadata file, it becomes difficult to store state beyond a couple of main attributes. I have gone ahead and implemented the ```get_model_args``` function showing how we would do it for ```WaveformModel```, but again, it only stores a subset of the state as many of the internal attributes can't be directly stored. 

The old tests have now gone, and new unit tests will need to be added for ```WaveformModel``` saving to check it all works as expected. But in I just wanted to ensure we are both on the same page before continuing. 

I've had a further look into the PyTorch Lightning approach, and I think long-term, this is definitely the most feasible option. Let me know your thoughts!

seisbench/models/base.py
```diff
@@ -192,6 +193,135 @@ def download_metadata_callback(metadata_path):
...
+
+        model = cls(**model_args)
+        model._weights_metadata = weights_metadata
+        model._parse_metadata()
+
+        model.load_state_dict(model_weights)
+
+        return model
+
+    def save(self, path, document=""):
```
yetinam, line 230: For consistency, I'd stick with `weights_docstring` instead of `document`.

seisbench/models/base.py
```diff
@@ -213,6 +344,23 @@ def _parse_metadata(self):
         default_args = self._weights_metadata.get("default_args", {})
         self.default_args.update(default_args)
 
+    def _get_input_args(self, obj):
```
yetinam, line 347: Could we move this function to `test_models.py` as a helper? I think it's not used anywhere except in the tests.

seisbench/models/aepicker.py
```diff
@@ -127,3 +127,27 @@ def classify_aggregate(self, annotations, argdict):
             )
 
         return sorted(picks)
+
+    def get_model_args(self):
+        model_args = super().get_model_args()
+        model_args = {
```
yetinam, line 133: This code will definitely work, but maybe it would look a bit cleaner without the dict comprehension. How about something like:
```python
for key in ["citation", "in_samples", "output_type", "default_args", "pred_sample", "labels", "sampling_rate"]:
    del model_args[key]
```
I think this would be more explicit

seisbench/models/base.py
```diff
@@ -213,6 +344,23 @@ def _parse_metadata(self):
...
+        signature = inspect.signature(obj)
+        args = {k: v._default for k, v in signature.parameters.items()}
+        if "kwargs" in args.keys():
+            del args["kwargs"]
+        return args
+
+    @abstractmethod
+    def get_model_args(self):
```
yetinam, line 355: I think this should return `{"citation": self.citation}`

seisbench/models/base.py
```diff
@@ -1374,6 +1522,20 @@ def detections_from_annotations(annotations, threshold):
 
         return detections
 
+    def get_model_args(self):
```
yetinam, line 1525: I'd add a call to the super-function to also get the citation.

seisbench/models/aepicker.py
yetinam: Same for the other models.

tests/test_models.py
```diff
@@ -10,6 +10,7 @@
 import logging
 import pytest
 import asyncio
+import os
```
yetinam, line 13: Looks like the import is unused.

yetinam (changes requested):
Hey @jawooll ,

thanks a lot for the PR. I only have a few cosmetic suggestions, but everything works very nicely now and the interface is smooth.

Last thing to do would be to marry it with `from_pretrained` and the versioning. I can do this after you implemented (or rejected) the cosmetic changes. Should we do this in the same PR or make another one for it afterwards.

And I guess we'll have to rebase the branch but that hopefully shouldn't be too problematic.

seisbench/models/base.py
jawooll: Yeah makes complete sense. This was legacy code from the old approach but is now not required as you say, moved to tests.

seisbench/models/base.py
jawooll: Correct, missed the top level out and is now included and passed down the object hierarchy.

seisbench/models/base.py
jawooll: As above, well-spotted and now added in.

seisbench/models/aepicker.py
jawooll: Sounds good, syntax changed to the above format for parsing any model keys.

tests/test_models.py
jawooll: Ah just spotted this, I'll add one additional commit to just remove this non-required dependency as we are going to rebase anyhow.

jawooll:
Hey Jannes, thanks for the review to catch these final points. I've gone through and implemented your suggestions. 

I think we should rebase and merge now, and then make a separate PR for integrating saving with ```from_pretrained``` afterwards. Basically 'separation of concerns' but for our git PRs (each PR implements a single new features for easy reference). What do you think? If we go for this I'll go through and rebase with main.

yetinam (approved):
Looks good to me now! Feel free to rebase and merge. I'll create the PR for combining this with from_pretrained afterwards. And then I guess we are ready to call it v0.2.

jawooll:
Thanks again for the detailed review, all closed and merged now. 

In the future, we can turn to trying to do this automatically, and also expand on this to incorporate model-checkpointing during training.

PR #72, opened 2022-03-18
Example for dataset creation

### LINKED ISSUE #58: Any workflow or examples of building a new standard seisbench dataset?
I'm a student and I am very interested about this ML toolbox in seismology, but I am not very familiar with ML and python, I was wondering that is there any workflow or example of converting my own data to a new standard seisbench dataset? I have read the documentation and found some functions about this but still not understand clearily.
Thank you very much.

yetinam:
This PR adds an example notebook for creating a new dataset in SeisBench format. In addition, it adds the trace_ops to the `__init__.py` of seisbench.util.

Please do not squash the commits when merging, so that I can cherry pick the trace_ops commit to `maintenance_0.1.x`. This would be helpful, so that we can publish a 0.1.14 release which would already allow to run the notebook on Colab before we finish 0.2.0.

Closes #58


jawooll (commented):
Thanks for implementing this. Noted for the point about not squashing the commits. I've just spotted a couple of minor typos in the examples commentary so I'll just change them and then it should be good-to-go.

yetinam:
Thanks! I'll cherrypick the commit to main.

Issue #76, opened 2022-04-01
Add model weight versioning


yetinam:
As #73 nicely showed, it might sometimes be desirable to update models weights. I'd therefore suggest to add model weight versioning. My proposal would be to keep all versions available in the repo, but by default download the latest version. Older versions can explicitly be queried using a `version` kwarg. If a version is already cached locally, checks for new versions are only performed when **explicitly initiated** by the user, for example through an argument like `check_latest`. This way, we avoid that SeisBench needs to "call home" every time `from_pretrained` is called. I think that would be a privacy issue.

#49 proposes a similar addition for datasets. However, I think versioning for models is considerably easier as the file size is much smaller.

jawooll:
Just from my side, I think this is a good-idea. It also slightly relates to #71. We can actually integrate this into the model saving and loading once implemented, where some form of 'development' versioning can be defined for local saving. This allows for automatic model checkpointing during training.

PR #77, opened 2022-04-04
Implement model weight versioning (#76)

yetinam:
This commit adds versioning for model weights. This commit includes the following changes:
- Added a weights_version property to SeisBenchModel.
- Added a "version" field to json configs of model weights.
- json config files are now mandatory for each set of model weights (but using the standard pytorch interface, it is still possible to load weights without config).
- The local cache structure now stores models as [name].[json|pt].v[version], i.e., a version suffix was added. Old caches are automatically converted. For compatibility reasons, the required remote cache structure is still downward-compatible.
- The list_pretrained function now takes local models into account.
- Added list_versions function.
- Implemented fine-grained control whether the remote repository or the local cache should be used.
- The documentation has been updated to include a remark on versioning.

In addition, the control flow of these functions was modified. Tests for all changes were added.

### DIFF
--- a/seisbench/models/base.py
+++ b/seisbench/models/base.py
@@ -18,6 +18,7 @@ from obspy.signal.trigger import trigger_onset
 import asyncio
 import nest_asyncio
 from packaging import version
+import tempfile
 
 
 class SeisBenchModel(nn.Module):
@@ -32,6 +33,7 @@ class SeisBenchModel(nn.Module):
         super().__init__()
         self._citation = citation
         self._weights_docstring = None
+        self._weights_version = None
         self._weights_metadata = None
 
     def __str__(self):
@@ -53,6 +55,10 @@ class SeisBenchModel(nn.Module):
     def weights_docstring(self):
         return self._weights_docstring
 
+    @property
+    def weights_version(self):
+        return self._weights_version
+
     @classmethod
     def _model_path(cls):
         return Path(seisbench.cache_root, "models", cls._name_internal().lower())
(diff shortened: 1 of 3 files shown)


jawooll:
Thanks for the PR and the additions, I'll take a look through.

seisbench/models/base.py
```diff
@@ -62,73 +68,69 @@ def _remote_path(cls):
...
-
-        def download_weights_callback(weight_path):
-            seisbench.logger.info(f"Weight file {name} not in cache. Downloading...")
-            weight_path.parent.mkdir(exist_ok=True, parents=True)
+        cls._cleanup_local_repository()
+        if version_str == "latest":
+            versions = cls.list_versions(name, remote=update)
+            if len(versions) == 0:
+                raise ValueError(f"No version for weight '{name}' available.")
+            version_str = max(versions, key=version.parse)
```
jawooll, line 124: This is currently breaking on version lookup when trying to obtain any of the pre-trained models:


```python
ValueError                                Traceback (most recent call last)
<ipython-input-17-4adba3d07bee> in <module>
----> 1 gpd = sbm.PhaseNet.from_pretrained('stead')

~/seisbench-production/seisbench/seisbench/models/base.py in from_pretrained(cls, name, version_str, update, force, wait_for_file)
    121             versions = cls.list_versions(name, remote=update)
    122             if len(versions) == 0:
--> 123                 raise ValueError(f"No version for weight '{name}' available.")
    124             version_str = max(versions, key=version.parse)
    125 

ValueError: No version for weight 'stead' available.
```

```list_versions``` works as expected. So you get back a list of 'versions'. I think you need to change the version.parse function you pass to the ```map(...)``` call as I can't find the version variable in the method scope.

File: seisbench/models/base.py [hunk 2]
```diff
@@ -141,60 +143,257 @@ def download_metadata_callback(metadata_path):
...
+                with open(metadata_path, "r") as f:
+                    weights_metadata = json.load(f)
 
-                if metadata_path.is_file():
-                    with open(metadata_path, "r") as f:
-                        config = json.load(f)
-                    detail_weights[weight] = config.get("docstring", None)
-                else:
-                    detail_weights[weight] = None
+        return weights_metadata.get("docstring", None)
```
jawooll, line 299: Similar to the ```from_pretrained``` error. We see the following when trying to call the internal ```_get_latest_docstring``` method:

```python
MissingSchema                             Traceback (most recent call last)
<ipython-input-33-1d1dafe38e58> in <module>
----> 1 sbm.PhaseNet._get_latest_docstring('stead', True)

~/seisbench-production/seisbench/seisbench/models/base.py in _get_latest_docstring(cls, name, remote)
    291             with tempfile.TemporaryDirectory() as tmpdir:
    292                 metadata_path = Path(tmpdir) / "metadata.json"
--> 293                 util.download_http(
    294                     remote_metadata_path, metadata_path, progress_bar=False
    295                 )

~/seisbench-production/seisbench/seisbench/util/file.py in download_http(url, target, progress_bar, desc, precheck_timeout)
     29     seisbench.logger.info(f"Downloading file from {url} to {target}")
     30 
---> 31     precheck_url(url, timeout=precheck_timeout)
     32 
     33     req = requests.get(url, stream=True, headers={"User-Agent": "SeisBench"})

~/seisbench-production/seisbench/seisbench/util/file.py in precheck_url(url, timeout)
     75 
     76     try:
---> 77         req = requests.head(url, timeout=timeout, headers={"User-Agent": "SeisBench"})
     78 
     79         if req.status_code >= 400:  # Assumes all 200 and 300 codes are acceptable

/work/software/conda/2020.11/lib/python3.8/site-packages/requests/api.py in head(url, **kwargs)
    102 
    103     kwargs.setdefault('allow_redirects', False)
--> 104     return request('head', url, **kwargs)
    105 
    106 

/work/software/conda/2020.11/lib/python3.8/site-packages/requests/api.py in request(method, url, **kwargs)
     59     # cases, and look like a memory leak in others.
     60     with sessions.Session() as session:
---> 61         return session.request(method=method, url=url, **kwargs)
     62 
     63 

/work/software/conda/2020.11/lib/python3.8/site-packages/requests/sessions.py in request(self, method, url, params, data, headers, cookies, files, auth, timeout, allow_redirects, proxies, hooks, stream, verify, cert, json)
    514             hooks=hooks,
    515         )
--> 516         prep = self.prepare_request(req)
    517 
    518         proxies = proxies or {}

/work/software/conda/2020.11/lib/python3.8/site-packages/requests/sessions.py in prepare_request(self, request)
    447 
    448         p = PreparedRequest()
--> 449         p.prepare(
    450             method=request.method.upper(),
    451             url=request.url,

/work/software/conda/2020.11/lib/python3.8/site-packages/requests/models.py in prepare(self, method, url, headers, files, data, params, auth, cookies, hooks, json)
    312 
    313         self.prepare_method(method)
--> 314         self.prepare_url(url, params)
    315         self.prepare_headers(headers)
    316         self.prepare_cookies(cookies)

/work/software/conda/2020.11/lib/python3.8/site-packages/requests/models.py in prepare_url(self, url, params)
    386             error = error.format(to_native_string(url, 'utf8'))
    387 
--> 388             raise MissingSchema(error)
    389 
    390         if not host:

MissingSchema: Invalid URL 'stead.pt': No schema supplied. Perhaps you meant http://stead.pt/?
```

Again, perhaps Line #282 is the culprit, as I can't seem to find any version variable in scope.

seisbench/models/base.py
```diff
@@ -141,60 +143,257 @@ def download_metadata_callback(metadata_path):
...
+
+            versions = list(set(versions + remote_versions))
+
+        return sorted(versions)
+
+    @classmethod
+    def _get_version_of_remote_without_suffix(cls, name):
+        """
+        Gets the version in the remote config without version suffix. Assumes this config exists.
+        """
```
jawooll, line 366: This is slightly pedantic, but just adding in here for reference. The way the method and docstring are worded, calling ```_get_version_of_remote_without_suffix(...)``` is slightly confusing as the file suffix is __.v*__ so, strictly speaking, removing the suffix would be to remove all of this. As we get back the integer component of the version suffix, can we change to ```_get_version_number_only_from_remote``` or similar.

```python 
sbm.PhaseNet._get_version_of_remote_without_suffix('stead')

>>> '1'
```

seisbench/models/base.py
jawooll: Additional note: This only raises an error if the associated remote files for the pre-trained model have not been downloaded yet and a remote cache lookup is required. If the model (with versioning added) exists in the local cache, it works as expected (which is probably why the associated unit test for the method did not flag this).

jawooll (changes requested):
Hey Jannes, 
Thanks again for the PR. I've gone through the review and made some comments. Most of the functionality works as expected, as is perfectly documented and tested. The new changes do however break the ```from_pretrained``` calls. 

Simulating the remote connections are obviously difficult to flag in your unit tests, hence why this wasn't picked up. I add some examples of the errors raised and where I think these problems are coming from.

seisbench/models/base.py
yetinam: The issue here is that by default we use `update=False`. However, this also means that we are ignoring remote versions even if **no** local version is available. I changed the behaviour to include remote versions also if no local version is available.

seisbench/models/base.py
yetinam: The culprit is line 290, i.e., instead of generating an URL the code only generated the filename. I fixed this now.

seisbench/models/base.py
yetinam: Actually, the function does something slightly different. Assuming there is a file called `[name].json`, it returns the version reported in the json. This is required to figure out if this file might be the newest version. I've updated the docstring to make this more clear.

yetinam:
Hey Jack,
thanks for the review and spotting these issues. Quite bitter that the central function did not work anymore and the tests weren't catching it. If fixed the issues now and extended the tests. The new tests should have caught the issues. Looking forward to the second round of reviews.

jawooll (approved):
Thanks for implementing the fixes. This all seems to work now and integrates nicely! - merging now, as a quick heads-up, I'm planning on a 'squash and commit'. Is this ok with you, or would you prefer to keep the the distinct commits in this history?

yetinam:
Thanks for the review! I agree, squashing seems a good choice for this PR.

jawooll:
Ok great, all merged into main - cheers :)

PR #81, opened 2022-04-13
Implement parallel backend for annotate and extend documentation on parallelism

yetinam:
This PR adds a parallel backend for `WaveformModel.annotate`. However, as the overhead for starting processes in python is rather large, this would lead to large runtime increases when annotating short waveform fragments. Therefore, I implemented a dual system with two options to choose from for the user:

- the previous `asyncio` based implementation, which is particularly well suited for small inputs
- a new `multiprocessing` based implementation, which is aimed at annotating large inputs, where the overhead becomes small compared to the total runtime.

The amount of code required for implementing a new model stays unchanged.
In addition, this PR extends the documentation with hints on multiprocessing.

### DIFF
--- a/seisbench/models/base.py
+++ b/seisbench/models/base.py
@@ -1,5 +1,6 @@
 import seisbench
 import seisbench.util as util
+from seisbench.util import log_lifecycle
 
 from abc import abstractmethod, ABC
 from pathlib import Path
@@ -18,6 +19,51 @@ from obspy.signal.trigger import trigger_onset
 import asyncio
 import nest_asyncio
 from packaging import version
+import torch.multiprocessing as torchmp
+import logging
+
+
+@log_lifecycle(logging.DEBUG)
+def _watchdog(queue_watchdog, tasks):
+    """
+    Watchdog that terminates jobs once jobs in level before have been terminated.
+
+    :param queue_watchdog: Signal queue for watchdog
+    :param tasks: List of tasks.
+                  Each tasks consist of a 4-tuple: `key`, `target_queue`, `n_inputs`, `n_outputs`
+                  `key` is the key to track on the watchdog.
+                  `target_queue` is the queue to send `None` values to once condition is met.
+                  Can also be a list of queues. In this case, `n_outputs` stop commands are sent to each queue.
+                  `n_input` is the number of times `key` needs to be received before, i.e., the number of jobs that will
+                  send `key`.
+                  `n_outputs` is the number of `None` value to send to `target_queue`, i.e., the number of dependent
+                  jobs to terminate.
+    :return: None
+    """
+    task_counter = {
+        key: n_inputs for key, _, n_inputs, _ in tasks
+    }  # Count down how often a job existed
+    on_complete = {
+        key: (target_queue, n_outputs) for key, target_queue, _, n_outputs in tasks
+    }  # Action once condition is met
+
+    while True:
+        elem = queue_watchdog.get()
+        if elem is None:
+            break
+
+        task_counter[elem] -= 1
+        if task_counter[elem] == 0:
+            target_queues, n_outputs = on_complete[elem]
+            if not isinstance(target_queues, list):
+                target_queues = [target_queues]
+
+            for target_queue in target_queues:
+                target_queue.join()  # Makes sure everything in this queue is actually processed
+                for _ in range(n_outputs):
+                    target_queue.put(None)
+
+
 import tempfile
 
 
@@ -550,8 +596,20 @@ class WaveformModel(SeisBenchModel, ABC):
         self.labels = labels
 
         self._annotate_function_mapping = {
-            "point": (self._cut_fragments_point, self._reassemble_blocks_point),
-            "array": (self._cut_fragments_array, self._reassemble_blocks_array),
+            "point": (
+                (self._async_cut_fragments_point, self._async_reassemble_blocks_point),
+                (
+                    self._process_cut_fragments_point,
+                    self._process_reassemble_blocks_point,
+                ),
+            ),
+            "array": (
+                (self._async_cut_fragments_array, self._async_reassemble_blocks_array),
+                (
+                    self._process_cut_fragments_array,
+                    self._process_reassemble_blocks_array,
+                ),
+            ),
         }
 
         self._annotate_functions = self._annotate_function_mapping.get(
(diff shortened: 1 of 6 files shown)


jawooll:
Hey, thanks for this Jannes. This seems like it will be a really useful addition for potentially more intensive training pipelines. I'll take a look through.

seisbench/models/base.py
```diff
@@ -595,34 +683,67 @@ def annotate(
                                                This is usually acceptable for rotationally invariant models,
                                                e.g., most picking models.
         :type flexible_horizontal_components: bool
+        :param parallelism: If None, uses the `asyncio` implementation. Otherwise, defines the redundancy for each
+                            subjob, i.e., parallelism=2 would start each subjob twice. See the warning above for a
+                            discussion on parallelism for annotate.
+        :type parallelism: None, int
```
jawooll, line 689: 0 indexing works for this parameter though. So if I enter ```parallelism=0``` I'm still using the multiprocessing approach via ```self._annotate_processes(...)``` and I get back a reasonable input? I can't currently equate this with what's listed in the docstring (starting each subjob 0 times, but getting back something which looks reasonable). 

Just slightly confused with the difference between parallelism=0 and parallelism=1

seisbench/models/base.py
```diff
@@ -595,34 +683,67 @@ def annotate(
...
+        if parallelism is None:
+            nest_asyncio.apply()
+            call = self._annotate_async(
+                stream, strict, flexible_horizontal_components, **kwargs
+            )
+            return asyncio.run(call)
+        else:
+            return self._annotate_processes(
+                stream, strict, flexible_horizontal_components, **kwargs
+            )
```
jawooll, line 705: linked to the above comment, this is just showing that parallelism is allowed to be a value of 0, and in the control flow we then still use multiprocessing.

seisbench/models/base.py
```diff
@@ -595,34 +683,67 @@ def annotate(
...
+    def _check_parallelism_annotate(stream, parallelism, thresholds=(5e5, 5e7)):
+        """
+        Checks whether the chosen parallelism looks reasonable and prints a warning otherwise.
+
+        :param stream: Data stream
+        :type stream: obspy.Stream
+        :param parallelism: parallelism indicator
+        :type parallelism: None, int
+        :return: None
+        """
```
jawooll, line 717: I really like this helper method! As a quick general question from my side, were these thresholds chosen arbitrarily or is this where the runtime performance becomes quicker for multiprocessing over asyncio from some quick exploratory tests?

seisbench/models/base.py
```diff
@@ -722,12 +841,224 @@ async def _annotate_async(
...
+        for proc in (
+            cut_processes + pre_processes + post_processes + reassemble_processes
+        ):
+            proc.start()
+        process_predictions_to_streams.start()
+
+        for group in groups:
+            queue_groups.put(group)
+
+        process_watchdog.start()  # Why does it cause a race condition?
```
jawooll, line 1030: If this is still an issue, can we elaborate on this comment or tag it with a TODO:

seisbench/models/base.py
```diff
@@ -745,237 +1076,485 @@ async def _process_streams_to_arrays(
...
+
+    def _cut_fragments_array(self, elem, argdict):
+        """
+        Cuts numpy arrays into fragments for array prediction models.
+        """
+        overlap = argdict.get("overlap", 0)
+
+        t0, block, trace_stats = elem
+
+        bucket_id = np.random.randint(1000000)
```
jawooll, line 1364: We can boost this value all the way to 1e9 without affecting compute time, to lower probably of collisions. It is still probably not the optimum way of ensuring there are no collisions, but this approach provides the quickest way of assigning bucket-ids with v. low probability of the id matching.

jawooll (changes requested):
Hey, I've gone through and added my review. Again, this is a really nice, useful addition. The only minor sticking point I had was when testing the edge cases of the 'parallelism' parameter. I can specify ```parallelism=0' and get back something that is reasonable. So, are the jobs 0-indexed meaning that parallelism=2 in fact starts 3 subjobs, or does the control flow account for this case in addition to parallelism=1. In either case, this edge case would just have to be added to the docstring. 

The remaining additions all look good. Unit tests all pass and everything is well documented.

seisbench/models/base.py
yetinam: I think `parallelism=0` should be treated as `parallelism=None`. I'll adjust this.

seisbench/models/base.py
yetinam: I did a few quick tests to choose these values. But as you see, I added quite a big margin in between the lower and upper bound. I think the actual break-even points will depend on a lot of factors, e.g., the operating system or the model. For example, DeepDenoiser has pretty expensive pre-/post-processing (FFTs). Consequently, for DeepDenoiser using parallelism pays of by far earlier than, for example, for PhaseNet.

seisbench/models/base.py
yetinam: Upps, left that in here. No, the issue was resolve by adding the `target_queue.join()` in the watchdog process. I'll remove it.

seisbench/models/base.py
yetinam: I'm fine with the value either way. I guess we are anyhow so close to uniformity over the workers (even if people go up to something like 100) that it won't really matter. Alternatively, one could generate the `bucket_id` by hashing some key, but there's not really an advantage in this and on the downside one needs to identify a good key to hash.

seisbench/models/base.py
jawooll: Yeah true - good point. I think as a general guide this is a great addition as a reminder/pointer that we have this functionality now available.

yetinam:
Thinking about it, it should have deadlocked with `parallelism=0`. Turns out I didn't actually pass the parallelism value to `_annotate_processes`. Added this now but also made sure that 0 is treated just as None.

seisbench/models/base.py
yetinam: Increased the value to 1e9 now.

yetinam:
I added the changes now. To keep the commit history clean, I amended the previous commit rather than adding a new one. I also rebased the branch onto main to avoid a merge commit.

jawooll:
Ok makes sense, cheers, I'll take a look through.

jawooll (approved):
Thanks for the additional amendments. Case 'parallelism=0' is now treated as 'None' (so using asyncio) and the rest of the comments are also nicely addressed - cheers!

Merging now.

yetinam:
Thanks for the quick and through review!

PR #85, opened 2022-04-26
Suppress warning for mixed dtypes when loading metadata

yetinam:
This PR suppresses the pandas warning for mixed dtypes when loading the metadata from csv. It occurs primarily because we allow empty metadata fields. However, the warning can be pretty annoying, in particular, when loading datasets with many chunks because each chunk will trigger a warning. At the same time, we cannot pass a dictionary with dtypes to `read_csv` due to the flexible names of metadata columns. I think suppressing the warning should be fine here.

@jawooll : Just running this by you to see if you agree on hiding the warning.

### DIFF
--- a/seisbench/data/base.py
+++ b/seisbench/data/base.py
@@ -14,6 +14,7 @@ import scipy.signal
 import copy
 from collections import defaultdict
 from collections.abc import Iterable
+import warnings
 
 
 class WaveformDataset:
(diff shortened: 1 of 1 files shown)


jawooll:
Hey thanks for the heads-up, yeah I agree we can remove this.

jawooll (approved):
Looks good! Changed approved.

jawooll:
Cheers for tackling, this Jannes.

PR #86, opened 2022-05-04
Additions to model saving and loading

yetinam:
- Added versioning capabilities to save and load functions
- Refactored from_pretrain to use load function

Once we merged this PR, I think we can publish SeisBench v0.2.

### DIFF
--- a/seisbench/models/base.py
+++ b/seisbench/models/base.py
@@ -10,7 +10,6 @@ import torch.nn as nn
 import torch.nn.functional as F
 from collections import defaultdict
 from queue import PriorityQueue
-import inspect
 import json
 import math
 import numpy as np
@@ -180,18 +179,7 @@ class SeisBenchModel(nn.Module):
             name, version_str, weight_path, metadata_path, force, wait_for_file
         )
 
-        with open(metadata_path, "r") as f:
-            weights_metadata = json.load(f)
-
-        model_args = weights_metadata.get("model_args", {})
-        model = cls(**model_args)
-
-        model._weights_metadata = weights_metadata
-        model._parse_metadata()
-
-        model.load_state_dict(torch.load(weight_path))
-
-        return model
+        return cls.load(weight_path.with_name(name), version_str=version_str)
 
     @classmethod
     def _cleanup_local_repository(cls):
@@ -457,7 +445,7 @@ class SeisBenchModel(nn.Module):
         return sorted([config[prefix_len:] for config in configs])
 
     @classmethod
-    def load(cls, path):
+    def load(cls, path, version_str=None):
         """
         Load a SeisBench model from local path.
 
@@ -465,14 +453,12 @@ class SeisBenchModel(nn.Module):
 
         :param path: Define the path to the SeisBench model.
         :type path: pathlib.Path ot str
+        :param version_str: Version string of the model. If none, no version string is appended.
+        :type version_str: str, None
         :return: Model instance
         :rtype: SeisBenchModel
         """
-        if isinstance(path, str):
-            path = Path(path)
-
-        path_json = path.with_name(path.stem + ".json")
-        path_pt = path.with_name(path.stem + ".pt")
+        path_json, path_pt = cls._get_weights_file_paths(path, version_str)
 
         # Load model metadata
         with open(path_json, "r") as f:
@@ -490,7 +476,7 @@ class SeisBenchModel(nn.Module):
 
         return model
 
-    def save(self, path, weights_docstring=""):
+    def save(self, path, weights_docstring="", version_str=None):
         """
         Save a SeisBench model locally.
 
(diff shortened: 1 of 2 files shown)


jawooll:
Hey Jannes, thanks for tagging the additions onto the model saving and loading, I'll take a look through now.

seisbench/models/base.py
```diff
@@ -457,22 +445,23 @@ def _get_versions_from_files(name, files):
...
 
         For more information on the SeisBench model format see py:func:`save`.
 
         :param path: Define the path to the SeisBench model.
         :type path: pathlib.Path ot str
+        :param version_str: Version string of the model. If none, no version string is appended.
+                            The version string can alternatively be provided by directly providing it with the `path`,
+                            e.g., my_weight.json.v1. The `version_str` will take precedence over the `path` if both
+                            indicate a version.
+        :type version_str: str, None
```
jawooll, line 460: Slightly confused about the docstring for this and what it indicates (from my understanding anyway I could be missing something!). 

As the ```path``` parameter is typically the base path and suffixes are then added to this, what would be the point in saving out a .json.v1 file only. If you can explicitly define the full path including version strings, then you have to also inlcude whether you are saving out a .pt or .json file. From my side, it would just be cleaner to always have the user define any ```version_str``` to load. Otherwise if defaults to the standard lookup (without a version suffix).

seisbench/models/base.py
```diff
@@ -512,12 +504,13 @@ def save(self, path, weights_docstring=""):
         :type path: pathlib.Path or str
         :param weights_docstring: Documentation for the model weights (training details, author etc.)
         :type weights_docstring: str, default to ''
+        :param version_str: Version string of the model. If none, no version string is appended.
+                            The version string can alternatively be provided by directly providing it with the `path`,
+                            e.g., my_weight.json.v1. The `version_str` will take precedence over the `path` if both
+                            indicate a version.
+        :type version_str: str, None
```
jawooll, line 511: As above.

seisbench/models/base.py
```diff
@@ -586,6 +579,38 @@ def _contains_callable_recursive(dict_obj):
 
         seisbench.logger.debug(f"Saved {self.name} model at {path.absolute}")
```
jawooll, line 580: This raises an error on string paths currently.

```python
AttributeError: 'str' object has no attribute 'absolute'
```

 I know we usually default to ```Path``` objects for everything but for ease-of-use during model saving and loading I would allow for this to be done using strings too which are then converted to Path objs. So just have the DEBUG line log the absolute path for bath cases accounting for whether the path parameter was a string or path object.

seisbench/models/base.py
```diff
@@ -586,6 +579,38 @@ def _contains_callable_recursive(dict_obj):
...
+            if ".json.v" in path.name:
+                p = path.name.find(".json.v")
+                version_suffix = path.name[p + 5 :]
+            elif ".pt.v" in path.name:
+                p = path.name.find(".pt.v")
+                version_suffix = path.name[p + 3 :]
+            else:
+                version_suffix = ""
+        else:
+            version_suffix = f".v{version_str}"
```
jawooll, line 600: This follows on from the docstring commentary. 

We pass through a single path parameter, and here we are checking if the path param also includes one of the two file types (.json, .pt). If it does, we append an appropriate version suffix but we do this for both the json and pt paths. 

So, users have to specify a single file type with extension (e.g. path="mymodel.json.v1") in order to save out and load a versioned model just from the path parameter, this edge case just seems slightly non-intuitive.

jawooll (changes requested):
Hey Jannes, 
Thanks for the additions! Overall, most things are really nicely integrated. I just have some comments surrounding the logic to parse versions directly from the path parameter. Just see the minor review comments related to this.

seisbench/models/base.py
jawooll: From my side, I would go with always using version strings, and then if none are provided, just defaulting back to looking up the base model with no automatic inference of the version suffix

yetinam:
Hey Jack, I admit, it's not very intuitive on the first glance. However, the key reason for me to implement this behaviour is IPython autocomplete (i.e. completion by hitting tab). If people can not pass in a file name (no matter which one), they can not use autocomplete to type in the path, which I assume people would like to do. So I think this is actually quite handy. What's your opinion?

jawooll:
Hey, Ah this explains it! 

Ok so, this actually is a big 'pro' for keeping it as is. My only issue is that, for the user, it will be quite hard to decipher what is being saved if we are specifying the full paths including file names. I would go with just specifying a base path in all ```path``` parameters, and add the ```version_str``` to this if specified. I know this would be a nice thing to have but this approach is simpler to those working with SeisBench models. If you are set on keeping in we would have to remedy the docstring to explicitly explain what happens when you want to specify the version in ```path``` parameter:

```python
Examples
--------
>>> model.save(path"my_model.json.v1")
...
This will save ...
```

yetinam:
Good point. I'll switch it to only specifying the base path as you suggested. I also noticed that if a folder contains the files ?`my_weight.json` and `my_weight.pt` autocomplete will anyhow conveniently stop at `my_weight`.

seisbench/models/base.py
yetinam: Fixed that. Interesting way to introduce errors by factoring out a function. Great catch!

yetinam:
I adjusted the implementation accordingly now. You can have another look. (I amended the commit and force-pushed, so there's still only one commit in the PR.)

jawooll (approved):
Thanks for the changes, looks good to me! Descriptions in the docstrings also nicely highlight the functionality. Good to merge I think.

jawooll:
Cheers for putting in these final additions - good for v0.2 now

yetinam:
Great! I'll create the release later today.

Issue #82, opened 2022-04-13
Requesting example on fine tuning a model


JuanJose9611:
Hi, i want to start off  by thanking you for creating and making available this  amazing tool.

I am currently working on a project and, initially, the idea was to train the EQTransformer model using data from our region of study (Jalisco, México), however after further research, the amount of data that was available seemed too small to properly train the model from the start so it was proposed to instead use transfer learning and fine tune a pre-trained EQT model. I have read both the original seisbench paper and "Wich picker fits my data?" and both seems to suggest that it is possible to do so. 

One of the example colab notebooks has an example of training PhaseNet using one of the available datasets, i was wondering if the procedure for fine tuning a model was the same as in this notebook except  using `from_pretrained` when first defining the model, or if the procedure was different.

If different:

- I was wondering if there was a resource or example of fine tuning a model using seisbench that you could share, as the only work ive done with deep learning on python was with computer vision on volcano monitoring so the tools available where a bit different.

Sorry for my bad english and thank you in advance for looking at my request.

Juan.

yetinam:
Hey Juan,

thanks for your interest in SeisBench! Indeed, you for running transfer learning experiments, the only thing you need to change is to use `from_pretrained` for model initialisation. Otherwise, you can 1:1 use the steps from the example notebook.

You might also want to have a look at the codes for our benchmark paper: https://github.com/seisbench/pick-benchmark
While there is no transfer learning here (but you can use the same trick with `from_pretrained`), it shows one option to set up a bigger training.

I hope this answers your question. I'm closing this issue for now, but feel free to reopen if you have any further questions.

Jannes

JuanJose9611:
Thank you for your response.

I have another question however i don't know how to reopen the issue or if i should make a new one so for now i will post it here.

Unfortunately the original idea for my thesis project had to be scraped because at the moment we do not have access to some of the data. This due to the fact that the stations that were going to be used belong to different institutions and some permisions were not given. 
Instead, we will make a catalog of seismic activity in the region using EQT and compare it with catalogs from other institutions, this idea was proposed because as of now there are several discrepancies in detections between institutions.

So far i have been playing around with one station streams  making annotation/classification with the EQT model. 
If i use more than one station per stream, as i understand, seisbench automatically separates by stations so that would be a non issue. When i use only one station and one event is detected i get one dictionary containing the detection, my question is:

- If i have two stations and there is an earthquake that is pressent on both stations, will seisbench give the two detections separately or only one?
[...]

PR #87, opened 2022-05-09
Example notebook for catalog generation

yetinam:
Hey @jawooll ,
I though it would be nice to have an example on how to create a catalog for SeisBench using some associator. And as I anyhow wanted to play around with phase association a bit, I took the chance and create the example using GaMMA. I think this will also be helpful for people with similar issues as in #82. Could you have a look over the PR and comment what you think about the example? As study region I went to my canonical example, i.e., Iquique aftershocks.


jawooll:
Hey Jannes, I like the idea! I'll take a look.

yetinam:
I pushed a new version, now only taking 12 hours of data it. Tested it on Colab with and without GPU. The longest waiting time (2 or 3 minutes) was actually for downloading the data from GEOFON.

jawooll:
Ok, it looks good to me. As we can't do the usual line-by-line highlighting as it is a notebook, I'll just post some additional suggestions here:

1. When creating the pick_df, does the phase have to be a lowercase string for input into GaMMa? If not, would it be possible to keep this capitalized?
2. Just close off the epicentral map plot with plt.show()
3. For the final plot of the waveforms, we could actually plot the y-axis as source-station distance - to back up your commentary - once you select an event, compute the distance to each station, this will be in the same row-order as the station_df, so you can then just use this as the y-axis instead of the Northing. 

The rest looks good!

jawooll (changes requested):
See comment above.

yetinam:
For 1., yes, unfortunately this is necessary for GaMMA. I'll address the other points. Good idea with the source-station distance!

yetinam:
I pushed a new version incorporating the suggested changes (2 and 3). Could you have a look?

jawooll:
Hey, thanks for the quick turnaround, it's basically there now. The final axis label just needs to be changed too to distance now instead of northing. 

If you would like to, we can automate the cuda backend check:

```python
if torch.cuda.is_available():
    picker.cuda()
```

But these are now both very minor, after this I say we are good to merge.

yetinam:
Sounds good! I'll implement the changes tomorrow.

yetinam:
I integrated the changes now.

jawooll:
All looks good, thanks for this. Merging now.

yetinam:
Thanks for the thorough reviewing!

PR #92, opened 2022-06-02
Adding support for component_orders like Z12 instead of ZNE.

borthom:
Former behaviour treats digits as integers which raises exceptions. For example, if there is only one component `'1'` component and all the others are missing, string operations as index() no longer work on `1` as integer.

These code adjustments ensure both `trace_component_order` and each of its component names are always interpreted as string.


seisbench/data/base.py
```diff
@@ -442,7 +442,7 @@ def _get_component_mapping(self, source, target):
         :param target:
         :return:
         """
-        source = list(source)
+        source = list(str(source))
```
yetinam, line 445: I'm not sure about this change. Why is it required? I think the other change should already encompass this option. It also breaks the possibility to use a list as source.

yetinam:
Hey @borthom ,
nice catch! I added a small comment to the code. Could you have a look at it? In addition, I have two more queries:
- First, every change fixing a bug should have a unit test. The test should fail without the fix, so that it ensures the bug has been fixed. Could you add such a test? Probably the easiest option is just to write out a small dataset with random data and read it back in.
- Second, please adjust the formatting. Just install the pre-commit hook and it will be fixed in the next commit.

There are a bunch of unrelated test failures caused by networking. Don't worry about those.

borthom:
I don't know why black_action fails even though it is installed in my local environment. Executing `python -m black . --check` on the seisbench folder, it tells me everything was fine:
Bildschirmfoto von 2022-06-16 20-13-28

seisbench/data/base.py
borthom: You're actually right. But pandas treats empty strings as NaN and therefore as floats. I added a type check that raises a ValueError in such a case. But it could also be taken care of directly after reading the metadata file, or even not at all since it is rather an error of pre-processing. I'd leave this decision to you.

seisbench/data/base.py
```diff
@@ -1,3 +1,5 @@
+import math
```
yetinam, line 1: This seems to be an unused import. Could you remove it?

seisbench/data/base.py
```diff
@@ -442,6 +448,17 @@ def _get_component_mapping(self, source, target):
...
+
+        if (
+            isinstance(source, float)
+            and np.isnan(source)
+            or (
+                (isinstance(source, list) or isinstance(source, str))
+                and not len(source)
+            )
+        ):
+            raise ValueError(f"Component order not set for trace or dataset.")
```
yetinam, line 460: I think this is a good way of going about this. I'd just slightly rephrase the exception text, as this will not be called on specific traces but only once the dataset is loaded (or the target component order is changed, but the exception should never happen then). How about "Component order not set for (parts of) the dataset."?

tests/test_component_mapping_fix.py
```diff
@@ -0,0 +1,49 @@
+import math
+import pathlib
```
yetinam, line 2: Just for consistency, could you replace this with `from pathlib import Path`

tests/test_component_mapping_fix.py
```diff
@@ -0,0 +1,49 @@
+import math
```
yetinam, line 1: Please merge this test into `test_data.py` instead. Currently, there is only one file with tests for every module. I guess we should change that on the long run, but until then I'd stick to the old scheme.

yetinam (changes requested):
Hey @borthom ,
thanks for the update. Except the black issue, I only have 4 very minor comments that should be quick to address. I think the PR is really nicely done and the test is very comprehensive.

yetinam:
Regarding the black issue, it might be that your black version is wrong. I reformatted the files locally with mine and it reverted all changes your version introduced. The version should be `22.3.0`. We had to update it a few weeks back due to a bug, so maybe this is where the issue comes from. However, I'm surprised the pre-commit did not automatically sort things out for you. If the issue persists, I can also impose the correct format on my side before merging. I'll just wait until you implemented the final changes.

borthom:
> Regarding the black issue, it might be that your black version is wrong. I reformatted the files locally with mine and it reverted all changes your version introduced. The version should be `22.3.0`. 

Thanks for the advise! My black version, in fact, was 19.10b0 as currently provided by the conda default channel, I switched to installing black with pip instead (alternatively, using the community based conda-forge channel would have worked as well).

yetinam:
Hi @borthom ,
the PR looks good to me now. Only remaining change is the unused `math` import. Could you delete that and then I'll merge the PR.

yetinam:
Thanks once more for the addition!

PR #97, opened 2022-06-09
Changed severity of mismatching traces in annotate from Exception to Warning (Closes #90)

### LINKED ISSUE #90: Change mismatching streams severity from exception to warning
The `annotate` function currently returns an exception, if for one time window, there are multiple input streams with mismatching values. Crucially, this will stop the annotation process for all elements of the stream. I'd suggest to instead log a warning and remove only the conflicting traces (all of them) from the stream and still annotate the rest.

@jawooll : What do you think about the change?

https://github.com/seisbench/seisbench/blob/136842a4ffbcac287d866cb917b01a81acf6c0b7/seisbench/models/base.py#L1886-L1889

yetinam:
See #90 for details

Technically, this is a change in behaviour and should therefore be in a new minor version instead of a new patch version. However, I'd suggest to just stick to the liberal interpretation of versioning we are currently using, given that we are still at major version 0, i.e., early development.

### DIFF
--- a/seisbench/models/base.py
+++ b/seisbench/models/base.py
@@ -914,7 +914,7 @@ class WaveformModel(SeisBenchModel, ABC):
         self.annotate_stream_pre(stream, argdict)
 
         # Validate stream
-        self.annotate_stream_validate(stream, argdict)
+        stream = self.annotate_stream_validate(stream, argdict)
 
         # Group stream
         groups = self.group_stream(stream)
@@ -1032,7 +1032,7 @@ class WaveformModel(SeisBenchModel, ABC):
         self.annotate_stream_pre(stream, argdict)
 
         # Validate stream
-        self.annotate_stream_validate(stream, argdict)
+        stream = self.annotate_stream_validate(stream, argdict)
 
         # Group stream
         groups = self.group_stream(stream)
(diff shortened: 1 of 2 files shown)


jawooll (changes requested):
Hey Jannes, thanks for the PR, just took a look through now and everything appears extensively tested and perfectly integrated 👍 . My only point to raise for discussion would be the use of the term 'mismatching' as strictly, we could say the _'No intersection, no matching'_ case i.e. a stream which has multiple segmented records of the same traces not overlapping in time (such as the one I outline below), is a 'mismatched' stream - but we don't do anything here.

image
Please just ignore the function change for the data in the unit tests, as this was to check everything was contiguous and stays in the correct order. 
 
I think the fact this method is for parsing mismatched overlaps is nicely explained in the docstring and in the warning itself, I would just change the method name to ```sanitize_mismatching_overlapped_records``` so things are slightly more explicit in terms of what this method does. Very minor pedantic point more for discussion, but apart from that all good to merge!

yetinam:
Thanks for the review. I renamed the function and changed the comments in the test to "No intersection same value" and "No intersection different value".

jawooll (approved):
Thanks for the minor edits, tests all pass w. everything nicely integrated, I'll go ahead and merge now.

yetinam:
Thanks! I'll cherry-pick it over to main.

PR #102, opened 2022-06-21
EQTransformer model conversion and bug fixes.

calum-chamberlain:
This PR replaces the previous EQTransformer model conversion notebook and adds a notebook for the conversion of the non-conservative model ("original"). Weight outputs from these notebooks are here:
- [original](https://www.dropbox.com/s/70dg6tf3rfawpi6/original2022.pt?dl=0)
- [conservative](https://www.dropbox.com/s/5vpr6mn8t19avxm/conservative2022.pt?dl=0)

This PR also fixes two minor differences between the seisbench and original EQTransformer implementations as identified in #96. They are:
- EQTransformer used an `epsilon` of `1e-7` for numerical stability, but seisbench defaults to `1e-5` - now `original_compatible` models use `eps=1e-7`.
- In the computation of softmax for self attention the numerator was normalized by the input rather than the emissions. This resulted in differences in predictions in some cases.


yetinam:
Hey Calum, thanks for submitting the PR! I'll look through it tomorrow. In the meantime, I'd like to ask your opinion on the naming of the model weights. In general, I think given the new naming scheme in the original implementation, the canonical choice would be "original" and "conservative". However, given that we already use "original" to refer to the weights now named "conservative", this will most likely lead to confusions. I was considering "original" and "original_nonconservative", but maybe you have a better idea here?

yetinam:
The PR looks good to me, I'll merge now. Thanks @calum-chamberlain for all the work put in! I'll push this to pypi as v0.2.3 and update the weights in our repository afterwards to make sure the version requirement can actually be fulfilled.

yetinam:
The weights are in the repository now: https://dcache-demo.desy.de:2443/Helmholtz/HelmholtzAI/SeisBench/models/eqtransformer/

Issue #89, opened 2022-05-11
Passing iterators to annotate/classify


yetinam:
Currently `WaveformModel.annotate` and `~.classify` get obspy stream objects as input. This however requires to load all data into memory before calling the functions, which will not be feasible for large datasets. To accommodate this option, we could allow passing an iterator to the functions, that sequentially returns waveform streams. Similarly, to mitigate the same issue for the output, we could allow to pass a callback to the output that directly consumes the annotations. For `classify`, we could do an "early condensation" of the annotations, i.e., only keep the picks.

```python
waveform_iterator = sbu.WaveformFileIterator(path, suffix="mseed")

def write_annotations_callback(annotations):
    annotations.write("some_path.mseed")  # Some path needs to be derived from annotations

model.annotate(waveform_iterator, callback=write_annotations_callback)
```

In addition, this would allow to operator models in a server mode, where the iterator simply waits for data until new data is provided.

I won't immediately work on this, but wanted to get your opinion on this ( @jawooll ) and save the idea somewhere.

yetinam:
Closing this in favor of #248 and #251. With the asyncio interface publicly available, there's not really a point to come up with a specific interface for the generator and (potentially) callbacks for the output.

PR #114, opened 2022-07-20
Speedup nan trim

calum-chamberlain:
This is a simple speed-up enhancement that speeds-up nan-trimming by about 4x on my machine. Trimming nan's at the start and end of predictions was one of the slower parts of the model workflow (taking about 0.15s for a day of 100 Hz data) - while this isn't that slow, this method runs in ~0.04s for the same data).

### DIFF
--- a/seisbench/models/base.py
+++ b/seisbench/models/base.py
@@ -1960,7 +1960,7 @@ class WaveformModel(SeisBenchModel, ABC):
         """
         mask = ~np.isnan(x)
         valid = np.nonzero(mask == True)[0]
-        mask[valid[0]:valid[-1]] = True
+        mask[valid[0] : valid[-1]] = True
         x = x[mask]
 
         return x, valid[0], len(x) - (1 + valid[-1])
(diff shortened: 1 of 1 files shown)


yetinam:
Hi @calum-chamberlain ,
thanks for the contribution. Always good to shave of a few unnecessary (split)seconds! PR looks good to me. Only small request: Could you remove the old method that you currently renamed to `_trim_nan_old`? I think this will improve the code health/clarity on the long run and it's still in the git history for reference.

calum-chamberlain:
Will do, sorry for such a minor contribution - I am working my way slowly through trying to get the GPU load more consistent and noticed this on my travels.

yetinam:
No worries, happy about small contributions in any case as well! And very much looking forward to your findings towards getting the GPU load more consistent. Managing to use resources more effective would be a major gain. The parallel interface (activated by setting `parallellism` to an integer) leads to certain improvements in the GPU utilization, but its still far from perfect. In addition, it only really pays of for large inputs, because of the overhead. I was already considering running the models in a sort of "server mode", i.e., starting all processes early and keeping them running to avoid the overhead, but that's also not quite perfect. Another thought in this direction, also taking into account that the size of data passed to `annotate`/`classify` is limited by the main memory, is #89.

PR #127, opened 2022-09-26
allow the model to output logits when training

JUNZHU-SEIS:
Maybe calculate the loss between ''softmax(logits)'' and ''target'' in training is more numerically stable.


yetinam:
Hey, thanks for the PR! I agree, it seems like a good option to output logits where desired. I have two suggestions:

- To remove redundancy in the code, switch the `forward` function to use the new `logits` function internally.
- For consistency, I think it would be great to do add the same possibility for all models. Could you maybe do this? Otherwise I can also add these commits later.

Please also fix the formatting failure. Details on this can be found [here](https://github.com/seisbench/seisbench/blob/main/CONTRIBUTING.md).

JUNZHU-SEIS:
Thanks for your reply. I prefer the first suggestion as it won't introduce too much complexity for users who just predicts. I have fixed the formatting failure.

yetinam:
Sorry, I might have misformulated my point slightly. Was typing on my phone and the comment got a bit brief.

What I wanted to suggest was to change forward to the following to avoid duplication:
```python
def forward(x):
   return self.softmax(self.logits(x))
```

For the second part, I would just add similar options to GPD, EQTransformer, DPPDetect, BasicPhaseAE, and CRED, i.e., give users direct access to the logits as well. However, this doesn't need to be in this PR, I can also do it later. Just let me know.

JUNZHU-SEIS:
> Sorry, I might have misformulated my point slightly. Was typing on my phone and the comment got a bit brief.
> 
> What I wanted to suggest was to change forward to the following to avoid duplication:
> 
> ```python
> def forward(x):
>    return self.softmax(self.logits(x))
> ```
> 
> For the second part, I would just add similar options to GPD, EQTransformer, DPPDetect, BasicPhaseAE, and CRED, i.e., give users direct access to the logits as well. However, this doesn't need to be in this PR, I can also do it later. Just let me know.

I'm sorry for misunderstanding your suggestions. Your first suggestion is very constructive. And it requires to define an extra valuable to determine the application mode of the PhaseNet model. For example, in training and inference (or test) mode the model outputs logits and softmax(logits), respectively.
```python
   def __init__():
       self.mode = 'train' # or inference

    def forward(self, x):
        x_in = self.activation(self.in_bn(self.inc(x)))

        x1 = self.activation(self.bnd1(self.conv1(x_in)))
        x2 = self.activation(self.bnd2(self.conv2(x1)))
        x3 = self.activation(self.bnd3(self.conv3(x2)))
        x4 = self.activation(self.bnd4(self.conv4(x3)))

        x = torch.cat([self.activation(self.bnu1(self.up1(x4))), x3], dim=1)
        x = torch.cat([self.activation(self.bnu2(self.up2(x))), x2], dim=1)
        x = torch.cat([self.activation(self.bnu3(self.up3(x))), x1], dim=1)
        x = torch.cat([self.activation(self.bnu4(self.up4(x))), x_in], dim=1)
        x = self.out(x)
        return (x if self.mode=='train' else self.softmax(x))
```

For the second part, I'm afraid I can't contribute more as those architectures are unfamiliar to me.

yetinam:
As we're aiming to put out SeisBench v0.3 in the near future and I think it would be good to have this PR in there, I took the liberty to do the last changes and rebase the branch onto main. I'll merge in the next minutes if the CI passes.

Thank you very much for the contribution!

PR #142, opened 2022-11-09
Add data grouping

yetinam:
This PR adds data grouping. This includes:

- grouping in WaveformDataset and MultiWaveformDataset
- a new GroupGenerator
- adjustment of several augmentations to deal with grouped data

Grouping allows to automatically retrieve sets of metadata and waveforms from a dataset. The waveforms are selected based on a common key. For example, this enables to group the data by event and then train multi-station models. However, as any grouping is possible, there are countless options for using the scheme.

This PR does not contain any models. It also does not contain any grouping for the `annotate`/`classify` function. Due to the scale of this feature addition, and to keep the PRs somewhat reasonably sized, I think it's still worthwhile including this part already.

### DIFF
--- a/seisbench/data/base.py
+++ b/seisbench/data/base.py
@@ -1094,9 +1094,17 @@ class WaveformDataset:
             metadata.append(trace_meta)
 
         if return_metadata:
-            return waveforms, metadata
+            return waveforms, WaveformDataset._pack_metadata(metadata)
         else:
             return waveforms
+
+    @staticmethod
+    def _pack_metadata(metadata):
+        """
+        Reformats a list of dict into a dict of lists. Assumes identical keys in all dicts!
+        """
+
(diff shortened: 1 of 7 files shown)


jawooll:
Hey Jannes, thanks for this! Looking good, and makes complete sense to add this incrementally. My time is pretty limited till the end of this week and I don't want to be potentially holding you up on my end with a response to the PR. If you do want to keep adding things go for it, I'll be able to take a look through early next week and just step through what the changes are.

jawooll (approved):
Hey, I've managed to have a look through now, stepping through locally and even though there are quite a few changes. Everything's implemented perfectly and the test coverage is extensive and passing. All looks good to merge from my side! Feel free to go ahead and do this (either squash, or keep the individual commits in the history for more granularity -whatever you prefer). 👍

PR #155, opened 2022-12-18
A torch version PhaseNet

JUNZHU-SEIS:
This model has the same architecture as the original PhaseNet model (Zhu et. al. 2018)

Issue #151, opened 2022-12-13
Is the PhaseNet architecture the same with the one in the original paper (Zhu et al, 2019) ?


zhong-yy:
This is the PhaseNet architecture from Zhu, W., & Beroza, G. C. (Geophysical Journal International, 2019, 216(1), 261-273.)
![Screenshot from 2022-12-13 21-34-21](https://user-images.githubusercontent.com/38816011/207336595-63c4d1a3-1a7b-46ec-aae5-7156be755acf.png)
Here at each depth, there are two layers. Following each down-sampling/up-sampling convolutional layer, there seems to be a 1x1 convolution layer that increases/shrinks the number of channels.

The following code is from the seisbench package (`seisbench/models/phasenet.py`). There is only one down-sampling/up-sampling layer at each depth. I am wondering if this implementation is equivalent to the above figure.
```python
    def __init__(
        self, in_channels=3, classes=3, phases="NPS", sampling_rate=100, **kwargs
    ):
        ...
        self.in_channels = in_channels
        self.classes = classes
        self.kernel_size = 7
        self.stride = 4
        self.activation = torch.relu

        self.inc = nn.Conv1d(self.in_channels, 8, 1)
        self.in_bn = nn.BatchNorm1d(8)

        self.conv1 = Conv1dSame(8, 11, self.kernel_size, self.stride)
        self.bnd1 = nn.BatchNorm1d(11)

        self.conv2 = Conv1dSame(11, 16, self.kernel_size, self.stride)
        self.bnd2 = nn.BatchNorm1d(16)

        self.conv3 = Conv1dSame(16, 22, self.kernel_size, self.stride)
        self.bnd3 = nn.BatchNorm1d(22)

        self.conv4 = Conv1dSame(22, 32, self.kernel_size, self.stride)
        self.bnd4 = nn.BatchNorm1d(32)

        self.up1 = nn.ConvTranspose1d(
            32, 22, self.kernel_size, self.stride, padding=self.conv4.padding
        )
        self.bnu1 = nn.BatchNorm1d(22)

        self.up2 = nn.ConvTranspose1d(
            44,
            16,
            self.kernel_size,
            self.stride,
            padding=self.conv3.padding,
            output_padding=1,
        )
        self.bnu2 = nn.BatchNorm1d(16)

        self.up3 = nn.ConvTranspose1d(
            32, 11, self.kernel_size, self.stride, padding=self.conv2.padding
        )
[...]

PR #158, opened 2023-01-05
Implemented original PhaseNet model and converted weights

yetinam:
As identified in #151 the previous implementation of PhaseNet in SeisBench was different from the original. This PR implements the original version and converts the original weights. The previous implementation of PhaseNet is kept around as PhaseNetLight for compatibility. This PR uses ideas from #155.


JUNZHU-SEIS:
This is quite an exciting update!
I have retrained the new PhaseNet model with DiTing, a data set from China. The new PhaseNet model exhibits less false picks on continuous data.
My question on the architecture of your new model is why you offset the skip connections with the convoluted outputs when concatenating them in the up-sampling layers. Is this related to PyTorch features? Any reply to this question is much appreciated.

yetinam:
Super interesting to see the retrained model and the differences to PhaseNetLight. That seems quite promising. I'll also go into more systematic tests over the next days.

Regarding the offsets, these were necessary to achieve binary compatibility with the original PhaseNet written in tensorflow. Padding works slightly differently in pytorch and tensorflow and this adjusts for it.

For the pretrained models, this is expected. I haven't uploaded any models yet because that would break current SeisBench installations. I've thought of a solution now and will fix this in the next days. Plus I still need to actually train those models but I'm on it.

yetinam:
The cache issue should be fixed now to. For the fixed PhaseNet there is only the original weight available so far though because I still need to train the others.

zhong-yy:
Hi, it seems that `sbm.PhaseNet.from_pretrained("original")` has a very bad performance. Are you sure it is the original weights?

yetinam:
Hi @zhong-yy ,
thanks for the observation. I'm sure for the weights themselves but maybe something in the processing around it does not currently fit? Could you please open a new issue for this (for easier tracking) and provide more details on the problems? In particular, it would be really helpful to have some examples.

zhong-yy:
@yetinam Thank you for your quick reply. I am sorry I may have misused the model. When I use it with ‘model.eval()’ the performance is bad. After I remove ‘model.eval()’ it works well.
But I don’t know when I should use model.eval(). I have opened a new issue about this question

PR #188, opened 2023-03-30
Fix amplitude normalization in PhaseNet(Light) and EQTransformer (Closes #187)

### LINKED ISSUE #187: Inconsistent normalization strategies
The data normalization strategies between annotate, the original implementation (and weights), and the benchmark derived weights is not always identical. This can lead performance degradation. The table below shows the strategies.

|               | Original | Benchmark | annotate |
|---------------|----------|-----------|----------|
| EQTransformer | std      | peak      | std      |
| PhaseNet      | std      | peak      | std      |
| GPD           | peak     | peak      | peak     |

To fix the issue, the choice of normalization in `annotate` should be made dynamic.

This was originally detected in #186 but is kept in this issue report for better tracking.

yetinam:
See linked issue for description.

I'll updated the corresponding weights in the repository while the CI is running.

### DIFF
--- a/seisbench/models/base.py
+++ b/seisbench/models/base.py
@@ -105,6 +105,13 @@ class SeisBenchModel(nn.Module):
     :type citation: str, optional
     """
 
+    # The model can list combination of weights and versions that should cause a warning.
+    # Each entry is a 3-tuple:
+    # - weights name regex
+    # - weights_version
+    # - warning message
+    _weight_warnings = []
+
     def __init__(self, citation=None):
         super().__init__()
         self._citation = citation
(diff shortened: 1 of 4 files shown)


yetinam:
Weights are updated in the repository now. Furthermore, there is now a warning if older weight versions with incorrect normalization are used. Waiting for the CI to pass and will merge then.

zhong-yy:
Hi @yetinam, loading the updated weight will give an error `ValueError: Weights require seisbench version at least 0.3.2, but the installed version is 0.3.1.dev11+g9b24b60.`

yetinam:
Hi @zhong-yy ,
This happens only on the dev version installed from source. Just run `pip install -e .` again in the repository after pulling the latest changes and it should run. (If it doesn't on main, you might have to run the command while on the maintenance branch. You can safely switch back to main after.)

zhong-yy:
Thank you. There are 3 maintenance branches; to be safe, I prefer to stay in main. I tried install from `pip install seisbench`. Although the version number is 0.3.2,  it seems not the latest code. Finally I just set `seisbench.__verison__="0.3.2"`.

PR #194, opened 2023-04-25
Adapted to inputs of different lengths

senli1073:
In the current version, `offset < 0` may occur in some cases.

### DIFF
--- a/seisbench/models/phasenet.py
+++ b/seisbench/models/phasenet.py
@@ -165,7 +165,13 @@ class PhaseNet(WaveformModel):
     @staticmethod
     def _merge_skip(skip, x):
         offset = (x.shape[-1] - skip.shape[-1]) // 2
-        x_resize = x[:, :, offset : offset + skip.shape[-1]]
+        if offset >=0 :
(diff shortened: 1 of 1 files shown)


yetinam:
Hi @senli1073 ,
thanks for your contribution. I'm currently at EGU so will only get to look at it next week. In the meantime, could you provide a slightly more detailed description of the rational for the PR?

senli1073:
> Hi @senli1073 , thanks for your contribution. I'm currently at EGU so will only get to look at it next week. In the meantime, could you provide a slightly more detailed description of the rational for the PR?

Hey @yetinam ,

In some cases, the last dimension of `x` may be smaller than `skip`, resulting in `offset < 0`. 

For example, when `in_samples = 6001`, the shape of `x` in the third upsampling stage is `[1, 16, 1500]`, but the shape of `skip` is `[1, 16, 1501]`, resulting in  `offset = (1500 - 1501) // 2 = -1`.

The method given by the PR is: when `skip.shape[-1] > x.shape[-1]`, pad `0.` to both sides of the last dimension of `x`.

PR #204, opened 2023-05-24
bug_fix: adding model.eval() and model.train() in example notebooks

mbagagli:
In the [`03a_training_phasenet.ipynb`](https://github.com/seisbench/seisbench/blob/main/examples/03a_training_phasenet.ipynb) notebook the `model.eval()` call was missing (code-block 11). This call is needed to close the re-trained model, and is done automatically by the `annotate` method. In the notebook, though, must be specified. I've added such a switch in the `test_loop` function as well.


yetinam:
Thanks for spotting and fixing the error!

PR #208, opened 2023-06-14
post-processing: removing bottleneck nanmean/max

miili:
A bottleneck is the usage of `nanmean` and `nanmax`.

This dirty quick fix is replacing those functions with fuctions from the library bottleneck (https://github.com/pydata/bottleneck).

For my usecase I sped up the inference by ~30%.

### DIFF
--- a/seisbench/models/base.py
+++ b/seisbench/models/base.py
@@ -11,6 +11,7 @@ from pathlib import Path
 from queue import PriorityQueue
 from urllib.parse import urljoin
 
+import bottleneck as bn
 import nest_asyncio
 import numpy as np
 import obspy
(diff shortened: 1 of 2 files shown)


seisbench/models/base.py
```diff
@@ -1829,6 +1830,9 @@ def _predict_buffer(self, buffer):
         finally:
             if train_mode:
                 self.train()
+        if torch.cuda.is_available():
+            torch.cuda.synchronize()
```
miili, line 1834: idk if this is in-line with the rest of the cuda handling.

The `synchronize` statement waits for all kernels to finish and helped profiling the inference call stack.

yetinam: I'm not sure if this synchronization is actually desired here. In general, torch will anyhow synchronize everything it needs right now. Does it have a performance advantage? If not, I'd suggest commenting it out and adding a note that this might be useful to add in when profiling the code.

miili: done

yetinam:
Thanks again for the contribution!

PR #214, opened 2023-07-02
asyncio: moving nest_asyncio to toplevel

miili:
Patching asyncio loop at top-level fixes this upstream "bug" https://github.com/erdewit/nest_asyncio/issues/22

There is no harm patching the loop before.

@yetinam What is the incentive to use `nest-asyncio`?


yetinam:
Hi @miili,

thanks for the contribution. Looks good, I'll have another look tomorrow and merge.

Regarding the question, `nest_asyncio` is used to make the code compatible with jupyter notebooks. The notebooks run their own asyncio loop so when SeisBench tries to create its own loop the code crashes. You can try it out by commenting out the `apply` call and trying to run the code in a notebook. However, I would like to get rid of `nest_asyncio` because it causes trouble with other libraries, e.g., dask. In principle, it should be possible to check if there is already an event loop running and to attach to that loop and only create a new loop if there is none running yet.

miili:
Maybe we should check whether the library is imported from a notebook. Following https://stackoverflow.com/a/22424821

yetinam:
Thanks, that looks like a good solution! I've added a few explanatory comments, so that it's possible to trace why this code is there in the future.

Needed to make two commits though because windows does not allow for a file being called `aux.py` 🤦

PR #222, opened 2023-08-04
Adding options for disabling normalization and noise label for ProbabilisticLabeller, and checking picks for AddGap and ChannelDropout.

zhong-yy:
## Changes
1. Create an bool-type variable "noise_column" in the PickLabeller class to determine whether a noise label will be constructed. If the noise_column is false, no noise label will be created, and labels will not be normalized.
2. Update the `SupervisedLabeller::_check_labels()` function
3. Update `test_probabilistic_pick_labeller()`

## A visual example
```python
from seisbench.generate import ProbabilisticLabeller
import numpy as np
import matplotlib.pyplot as plt
np.random.seed(42)
p=50
s=100
state_dict = {
    "X": (
        10 * np.random.rand(3, 200),
        {
            "trace_p_arrival_sample": p,
            "trace_s_arrival_sample": s,
        },
    )
}
fig,axs=plt.subplots(4,2,sharex="all")

for i,noise_column in enumerate([True,False]):
    labeller = ProbabilisticLabeller(
        dim=0, shape="gaussian",sigma=20, noise_column=noise_column
    )
    labeller(state_dict)

    state_dict["y"][0].shape
    print(np.all(state_dict["y"][0]>=0))

    for j in range(3):
        ax=axs[j][i]
        ax.plot(state_dict["X"][0][j,:].T)
    axs[3][i].plot(state_dict["y"][0].T)
    axs[3][i].axvline(p,color="blue",label="True P")
    axs[3][i].axvline(s,color="red",label="True S")
axs[3][1].legend()
```
image

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -6,8 +6,9 @@ repos:
     args: ['--maxkb=1000']
   - id: check-toml
   - id: check-json
-  - id: pretty-format-json
-    args: [--autofix]
+# Hook shows incompatibility with ipynb files
+#  - id: pretty-format-json
+#    args: [--autofix]
   - id: end-of-file-fixer
   - id: trailing-whitespace
 - repo: https://github.com/pycqa/isort
--- a/seisbench/generate/augmentation.py
+++ b/seisbench/generate/augmentation.py
@@ -395,7 +395,12 @@ class ChannelDropout:
                     # if all channels are zeros, ignore original phase arrivals
                     for key in metadata.keys():
                         if key.endswith("_arrival_sample"):
-                            metadata[key] = np.nan
+                            if isinstance(metadata[key], (int, np.integer, float)):
+                                # single window case
+                                metadata[key] = np.nan
+                            else:
+                                # multi-window case
+                                metadata[key] = [np.nan] * len(metadata[key])
 
         state_dict[self.key[1]] = (x, metadata)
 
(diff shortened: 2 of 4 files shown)


yetinam:
Thanks for the PR! I might take some days until I get around to review it.

zhong-yy:
Hi @yetinam, I also added some checks for the output of AddGap() and ChannelDropout(), see https://github.com/seisbench/seisbench/issues/223. 

If the waveform segment containing a phase is zeroed out after a gap is added, the corresponding phase arrival will be set to NaN. If all channels are zeros after dropping some channels, all arrival samples will be set to NaN.

yetinam:
Thanks for the addition. Please make the zeroing from ChannelDropout optional as well. I'll put an explanation in the issue thread. Could you also add two short unit tests for the new changes?

yetinam:
I'd suggest using np.allclose instead of np.isclose(np.sum())

zhong-yy:
Thank you for your feedback. I just add an option `check_picks_in_gap` for ChannelDropout. I  also changed the criterion for deciding whether a pick within a gap will be ignored. If the pick is within the gap and it is far enough from the border of the gap, then the pick is made NaN. 

I will add two unit tests tomorrow or in a few days.

yetinam (approved):
I've now reviewed the PR. Looks good and the test are very comprehensive! Thanks for the addition. I've added some very minor changes and will merge the PR once the CI passes.

Once again, thanks a lot for your contribution!

PR #229, opened 2023-08-14
Add an option for AddGap and ChannelDropout to modify label directly

zhong-yy:
In pull request https://github.com/seisbench/seisbench/pull/222, only the metadata is changed if there is any pick in a gap.

It may be helpful to add an option to change labels directly. For example. when one waveform is staked on another (e.g. [benchmark/augmentations.py](https://github.com/seisbench/pick-benchmark/blob/main/benchmark/augmentations.py)) to generate a new trace, the metadata is kept unchanged and only the label is changed. In this case, adding the gap to the labels directly is a better choice.

- I added two label_keys `label_keys` and `noise_id` to `AddGap` and `ChannelDropout`
- Two new tests: `test_channel_dropout_with_picks_in_gap` and `test_add_gap_to_waveform_and_labels`

I forgot to pull your commit at the beginning, so I manually resolved the conflicts. I hope it didn't break the commits b253784a084a10ed8af19950e780999e3118fd06 and 319d9447e71cf92033d5653560cb180c48209858

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -6,8 +6,9 @@ repos:
     args: ['--maxkb=1000']
   - id: check-toml
   - id: check-json
-  - id: pretty-format-json
-    args: [--autofix]
+# Hook shows incompatibility with ipynb files
+#  - id: pretty-format-json
+#    args: [--autofix]
   - id: end-of-file-fixer
   - id: trailing-whitespace
 - repo: https://github.com/pycqa/isort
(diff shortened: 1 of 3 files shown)


yetinam:
Thanks for the contribution. I'll have a look at it but may not be able to do so before next week.

yetinam (approved):
Looks all good 👍

yetinam:
Thanks for the contribution! That was a very pleasant PR to review.

Issue #220, opened 2023-08-01
Structural Modifications for Optimal Outputs of the classify method.


imankahbasi:
Hello,

I would like to propose a modification to the output structure of the "classify" method in the model. Specifically, when utilizing the EQTransformer model, I have observed that the current output of the classify method is in the form of a tuple containing two lists: detections and picks.

`results =  eqt.classify(st)`
which
`([list_of_picks], [list_of_detections]) == results`

However, when employing a different model, the result generated by the classify method is simply a list of picks.

`results =  pha.classify(st)`
which
`[list_of_picks] == results`

This scenario presents a challenge. To illustrate, I have a dataset and wish to evaluate three models. I achieve this by implementing a for-loop for the models. During this process, I need to create an if statement as follows:

```python
if model_name == 'eqt':
	picks, detections = model.classify(st)
else:
	picks = model.classify(st)
```

I recommend creating a class to represent the output of this method, such as:

```python
class output:
    def __init__(self):
        self.detections = []
        self. picks = []
```

The output class internally stores the results in these lists, and the `classify` method returns an instance of this class.
To access the detections and picks, simply use `results.detections` and `results.picks` respectively.
However, when using any model other than EQTransformer, `results.detections` will be an empty list.

This structure makes a standard behavior for all usages.

In addition, I encounter some challenges when attempting to print results. Therefore, we can augment the class by including the `__str__` method, which will streamline the process of printing out results.

```python
class output:
    def __init__(self):
         self.detections = []
         self.picks = []

    def __str__(self):
         txt = f'There are {len(self.detections)} numbers of detections:\n '
         for detect in self.detections:
             txt += str(detect) + '\n'
         txt = f'There are {len(self.picks)} numbers of picks:\n '
         for pick in self.picks:
             txt += str(pick) + '\n'
         return txt
```

We can also create a selection method similar to `stream.select` in the obspy package.

I greatly appreciate your excellent SeisBench package.
It's truly wonderful.

yetinam:
Hi @imankahbasi ,
thanks for the feature suggestion and the very comprehensive description. I think that's a great idea. There are a few more point to add from my side:

- To stay flexible for future models with other outputs, I'd suggest making the outputs a [namespace object](https://docs.python.org/3/library/types.html#types.SimpleNamespace). That'll also automatically give the __str__ method. Maybe it could be called ClassifyOutput.
[...]

PR #230, opened 2023-08-14
Unify classify structure

yetinam:
Implement the modifications to unify classify outputs suggested in #220. The two main changes are:

 - a unified `ClassifyOutput` object that's essentially a namespace with information on the creating model.
 - containers for picks and detections that allow easier filtering

See the issue for details.

**Only merge PR shortly before publishing v0.5 because it'll break the colab example that do not run on head!**


yetinam:
Adjust to #231 before merging

PR #240, opened 2023-09-26
Some improvenents/cleanups in the GEOFON module; added a docstring

jsaul:


seisbench/data/geofon.py
```diff
@@ -23,8 +24,65 @@ class GEOFON(BenchmarkDataset):
     GEOFON dataset consisting of both regional and teleseismic picks. Mostly contains P arrivals,
     but a few S arrivals are annotated as well. Contains data from 2010-2013. The dataset will be
     downloaded from the SeisBench repository on first usage.
+
```
yetinam, line 27: Thanks for adding the extensive docstring. I think it would be good to put a subheading in between to make clear these details refer to the source data for converting the dataset from scratch.

seisbench/data/geofon.py
```diff
@@ -23,8 +24,65 @@ class GEOFON(BenchmarkDataset):
...
+    In order to use the GEOFON dataset, in addition to the raw data as
+    described above, there needs to be an inventory file (as FDSN Station XML)
+    for all stream-time combinations. Data files for which no corresponding
+    inventory entries are found will be ignored. The inventory doesn't need to
+    contain instrument responses, as the latter are not needed. The inventory
+    file is currently expected as file with name
+    'inventory_without_response.xml'.
     """
 
+    start_train = "2010-01-01"
```
yetinam, line 82: Maybe we should also add an `end_data` field or similar to indicate data until which data was used in the conversion.

seisbench/data/geofon.py
```diff
@@ -156,6 +244,11 @@ def _get_event_params(event):
                 event_params["source_focal_mechanism_n_plunge"] = n_axis.plunge
                 event_params["source_focal_mechanism_n_length"] = n_axis.length
             except AttributeError:
+                print(
```
yetinam, line 247: Please use `seisbench.logger` instead of print.

seisbench/data/geofon.py
```diff
@@ -311,10 +404,19 @@ def fill_dicts(self):
...
+                    row["sensitivity"] is None
+                    or row["sensitivity"] == "None"
+                    or "sensitivity" not in row
+                ):
+                    sensitivity = np.nan
+                else:
+                    sensitivity = float(row["sensitivity"])
+            except:
+                print(row.keys())
+                raise
```
yetinam, line 418: I think this should be removed again to avoid crashes.

seisbench/data/geofon.py
```diff
@@ -343,3 +453,13 @@ def find(self, trace_id):
                 return self.short_dict[netsta]
             else:
                 return np.nan, np.nan, np.nan, np.nan
+
+
+class GEOFONv2(GEOFON):
```
yetinam, line 458: Before merging this into `main`, we should discuss if/when we will make the underlying dataset publicly available. In general, I would prefer not to put any dataset into `main` that are only private at GEOFON.

yetinam, line 458: Before merging this into `main`, we should discuss if/when we will make the underlying dataset publicly available. In general, I would prefer not to put any dataset into `main` that are only private at GEOFON.

yetinam:
PR is still a draft, do not merge yet!

seisbench/data/geofon.py
jsaul: That makes a lot of sense

seisbench/data/geofon.py
jsaul: Fully agreed

PR #244, opened 2023-11-05
DRAFT: asyncio: using threads

miili:
Moving heavy-duty GIL releasing work into threads.

Exposing `annotated_async`.

Updated pre-commit hooks.

### DIFF
--- a/seisbench/models/base.py
+++ b/seisbench/models/base.py
@@ -892,6 +892,7 @@ class WaveformModel(SeisBenchModel, ABC):
         self,
         stream,
         parallelism=None,
+        copy=True,
         **kwargs,
     ):
         """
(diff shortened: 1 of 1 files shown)


yetinam:
Hi @miili 
thanks for the contribution! This could really speed things up. I've cancelled the workflow from the first iteration as it seems to have run into an infinite loop. I'll have a detailed look at the PR in the next days,

miili:
Hi Jannes,

hope you are well!

I could not do detailed analysis and benchmark. However only when you kick the blocking function to a thread (`asyncio.to_thread`) the GIL is released to other workers in the event loop, i.e. numpy stuff.

In my brief check I had a performance decrease, but maybe you have a lean benchmark setup ready to test these minor changes.

Currently I am looking at a throughput of 40 - 45 MB/sec MiniSeed (~18 h data / sec) for the detections in lassie-v2, with the current bottleneck being the inference. 👍🏻

yetinam:
Okay, I finally got to play with this. I've added a very minimalist benchmark script but it does the job. I tested the PhaseNet performance (on CPU) and indeed, the modified version is slower (142 s) than the original version (132 s). I think the problem is that the threads would only help for IO limited tasks but not for compute limited tasks.

miili:
The CPU and GPU compute can run asynchronously and can improve the workflow. Yet only when the different pre- and post-processing is running from queue to queue _and_ CUDA is enabled.

Are the `asyncio.Queue`s size limited?

PR #271, opened 2024-02-08
base: stream copy now optional.

miili:
Adding argument copy: `annotate(st, copy=False)`

Safes me 0.5 seconds per call.

### DIFF
--- a/seisbench/models/base.py
+++ b/seisbench/models/base.py
@@ -892,6 +892,7 @@ class WaveformModel(SeisBenchModel, ABC):
         self,
         stream,
         parallelism=None,
+        copy=True,
         **kwargs,
     ):
         """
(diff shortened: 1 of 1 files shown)


yetinam:
Good idea! Assuming the CI passes, I'll merge this right away.

yetinam:
Just out of curiosity, how much data do you give per call and how long does the call take in total? Just for me to get a better understanding of the relative speed-up.

miili:
Yeah, sure.
With copy:
```
loading model for gpu 0
compiling model for gpu 0
processed in 0:00:07.741108
processed in 0:00:06.356077
processed in 0:00:06.332223
```

and without
```
loading model for gpu 0
compiling model for gpu 0
processed in 0:00:07.286087
processed in 0:00:05.955739
processed in 0:00:05.855024
```

it ain't much, but ~8%

PR #277, opened 2024-03-14
Bugfix for PhaseNet label IDs

JanisHe:
Hi Jannes,

when using the original PhaseNet model or diting as base models for transfer learning, the label IDs are wrong. Default labels_ids are {"P": 0, "S": 1, "Noise": 2}, however, PhaseNet and diting have {"Noise": 0, "P": 1, "S": 2}. Therefore, transfer learning was not possible since the ids did not depend on the given labels by the loaded model. I rewrote the method "_colums_to_dict_and_labels" and added an argument model_labels to get the correct dictionary and label IDs.

Cheers,
Janis

### DIFF
--- a/seisbench/generate/labeling.py
+++ b/seisbench/generate/labeling.py
@@ -143,7 +143,7 @@ class PickLabeller(SupervisedLabeller, ABC):
                 self.labels,
                 self.label_ids,
             ) = self._colums_to_dict_and_labels(
-                label_columns, noise_column=noise_column, model_labels=self.model_labels
+                label_columns, noise_column=noise_column, model_labels=model_labels
             )
         else:
             self.labels = None
--- a/seisbench/generate/labeling.py
+++ b/seisbench/generate/labeling.py
@@ -131,16 +131,19 @@ class PickLabeller(SupervisedLabeller, ABC):
     :param kwargs: Kwargs are passed to the SupervisedLabeller superclass
     """
 
-    def __init__(self, label_columns=None, noise_column=True, **kwargs):
+    def __init__(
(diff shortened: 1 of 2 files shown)


tests/test_generate.py
```diff
@@ -1051,22 +1051,32 @@ def check_labels_against_ids(labeller, ids, labels):
         labeller(state_dict)
 
 
-def test_colums_to_dict_and_labels():
+@pytest.mark.parametrize(
```
yetinam, line 1054: This test should also test the case that `model_labels` is None.

seisbench/generate/labeling.py
```diff
@@ -131,16 +131,19 @@ class PickLabeller(SupervisedLabeller, ABC):
     :param kwargs: Kwargs are passed to the SupervisedLabeller superclass
```
yetinam, line 131: Please add a docstring for `model_label`. And as I've just spotted that the one for `noise_column` is missing, would you mind adding that too?

yetinam:
Hi @JanisHe ,
thanks a lot for the contribution. Indeed, it's nice to add this flexibility here. I've left a few minor that I'd ask you to address.

I'm not sure I'd call it a bug but it's definitely good to add this functionality. However, that's why I'd prefer merging it into `main`. Would you mind switching the base branch and rebasing your PR?

JanisHe:
Hi Jannes,
I just add your points and then we can merge this to `main`

yetinam:
Hi @JanisHe ,
I'm looking at this now and modifying a few details. I'll hopefully merge it tomorrow.

yetinam:
I've merged the branch now. Thanks a lot for your contribution!

PR #280, opened 2024-03-18
Add warnings for use of default component order

jacksonfellows:
Warn the user when they don't specify a component order and the default value is used. Also warn again if the default component order doesn't match what is the dataset's data format.

Change tests that check for warnings to clear the message log after loading the dummy dataset, not before, so that the new warnings don't break any tests.

Example of what these warnings would look like for the PNW dataset:
```
2024-03-18 13:28:12,164 | seisbench | WARNING | Component order not specified, defaulting to 'ZNE'.
2024-03-18 13:28:12,164 | seisbench | WARNING | Default component order 'ZNE' does not match dataset component order 'ENZ'.
```


yetinam:
Hi @jacksonfellows ,
the PR looks good to me and I'm basically happy to merge. I'd just put it into the `main` branch directly instead of `maintenance`. Could you switch the base branch and rebase the changes. It'll be part of v0.7 that I hope to wrap up in the next few weeks.

yetinam:
I've merge the PR now. Thanks a lot for reporting the issue and providing a PR so quickly!

PR #282, opened 2024-03-30
Change annotate logic from windowed to batched pre- and postprocessing

### LINKED ISSUE #269: DeepDenoiser: Use torch for stft
The bottleneck in the DeepDenoiser module is the short time Fourier transform.
Can `scipy.signal.stft` be replace with `torch.stft` to profit from GPU?

https://pytorch.org/docs/stable/generated/torch.stft.html

The method args differ from SciPy and are not that feature-rich. But are there any roadblocks?

yetinam:
Pre- and postprocessing in the annotate function was so far done per window and in numpy. This PR changes the processing to a batched mode running on PyTorch. This also ensures that all processing can happen on the GPU if available. In addition, this now allows to wrap the three processing steps in a thread, allowing to release the GIL during the heavy numeric operations. Overall, this will reduce run times, in particular, on GPU.

As this change improves the efficiency of the asyncio implementation, it removes the (suboptimal) process implementation completely.

To ensure compatibility with the previous implementation, a suite of tests has been added and the former `annotate_window_pre`/`annotate_window_post` functions are retained. These functions and tests should be removed in a future PR.

Additional small changes:
- added minimal benchmark for annotate performance (results coming in a bit)
- increased default batch size from 64 to 256
- refactored some functions in seisbench.util

Closes #269

### DIFF
--- a/contrib/benchmark_model_annotate.py
+++ b/contrib/benchmark_model_annotate.py
@@ -1,0 +1,44 @@
+"""
+This is a minimalist benchmark to compare different versions of the annotate function
+"""
+
+import time
+
+import numpy as np
+import obspy
+
+import seisbench.models as sbm
+
+
+def main(n_stations: int = 2, n_samples: int = 8640000):
+    models = [sbm.PhaseNetLight(), sbm.PhaseNet(), sbm.EQTransformer()]
+    # models = [sbm.PhaseNet()]
+
+    data = obspy.Stream()
+    for sid in range(n_stations):
+        for c in "ZNE":
+            data.append(
+                obspy.Trace(
+                    np.random.rand(n_samples),
+                    header={
+                        "network": "CX",
+                        "station": f"S{sid:04d}",
+                        "location": "",
+                        "channel": f"HH{c}",
+                        "sampling_rate": 100.0,
+                    },
+                )
+            )
+
+    for model in models:
+        # model.cuda()
+        # Warmup
+        models[0].annotate(data[:3])
+        t0 = time.time()
+        model.annotate(data, batch_size=1024, copy=False)
+        t1 = time.time()
+        print(model.__class__.__name__, t1 - t0)
+
+
+if __name__ == "__main__":
+    main()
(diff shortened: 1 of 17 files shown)


yetinam:
Here are the benchmark results. All experiments were run on a server using 12 CPU cores and a V100 GPU for the GPU experiments. Times are in seconds to annotate 10 stations, each with 1 day of 3 component data at 100 Hz. The batch size was set to 4096 (even though the GPU could handle even more).

```
Old implementation:
CPU
PhaseNetLight 27.505552291870117
PhaseNet 34.15515637397766
EQTransformer 62.846691608428955
DeepDenoiser 515.2418355941772

GPU
PhaseNetLight 18.10126042366028
PhaseNet 18.107561111450195
EQTransformer 17.852076530456543
DeepDenoiser 394.84921383857727


New implementation:
CPU
PhaseNetLight 19.04576873779297
PhaseNet 25.985053300857544
EQTransformer 55.00685524940491
DeepDenoiser 160.62065076828003

GPU
PhaseNetLight 9.644033908843994
PhaseNet 9.980738639831543
EQTransformer 8.486823320388794
DeepDenoiser 17.370351314544678
```

Overall, there are speed-ups on both CPU and GPU in all configurations. The magnitude of speed-up vary a lot between the models, depending on how heavy pre- and postprocessing are compared to the model itself. A few quick observations:

- Speed-ups on CPU vary from 1.15x (EQTransformer) over 1.32x (PhaseNet) up to 3.2x (DeepDenoiser).
- Speed-ups on GPU are larger, varying from 1.81x (PhaseNet) over 2.09x (EQTransformer) up to 22.6x (DeepDenoiser).
- Consequently, speed-ups between CPU and GPU implementations have now increased. From 1.9x to 2.6x (PhaseNet), from 3.5x to 6.4x (EQTransformer), from 1.3x to 9.2x (DeepDenoiser).

Notably, PhaseNet, PhaseNetLight and EQTransformer can now annotate a day of data at 100 Hz on GPU in under a second, i.e., around 10e5 times faster than real time.

yetinam:
Remark: Dropping the process interface also made the tests much quicker.

PR #306, opened 2024-07-11
Add the VCSEIS dataset

zhong-yy:
Add a dataset of earthquakes from various volcanic regions.


yetinam:
Thanks for the PR and the contribution! I only did some very minor modification. The new dataset is now in main and will be released as part of v0.8.

Issue #130, opened 2022-09-30
Migration to type hints


jawooll:
With type hinting becoming both standardised and stable within the latest Python versions, maybe we should consider including type hints with the next minor release? As we support Python 3.7+ this is still ok. 

It would be a rather redundant addition, going through the entire codebase and implementing the appropriate changes, but will enhance readability and keep us better aligned with PEP guidelines.  

This can always be delayed to a later date too, just worth getting some thought's on this.

yetinam:
I agree, it seems like a good idea to replace the type hints in our documentation with proper type hints. Maybe more of a long term project to progressively address? I guess slowly adding type hinting module by module shouldn't break anything.

jawooll:
Definitely, doing it in the way you suggest will break this up into more manageable chunks - sounds good 👍

PR #313, opened 2024-08-06
New augmentation methods and bugfix in PhaseNet labels

JanisHe:
In #277 we forgot the case for "N", which is fixed with this pull request and also included in the tests.
I wrote two new augmentation methods to rotate three component data by an arbitrary angle and to add recorded noise from a previously loaded data set.


seisbench/generate/augmentation.py
```diff
@@ -3,6 +3,8 @@
 import numpy as np
 import scipy.signal
 
+import seisbench.data.base
```
yetinam, line 6: So far, the idea was that the three main modules of SeisBench do not import each other (https://github.com/seisbench/seisbench/blob/main/CONTRIBUTING.md#code-structure). From what I see, you only need it for the type hints here. I'd suggest removing the type hints for the WaveformDataset to get rid of the import. But maybe we should consider changing the policy and allow to import seisbench.data in seisbench.generate?

seisbench/generate/augmentation.py
```diff
@@ -670,3 +672,194 @@ def __call__(self, state_dict):
...
     def __str__(self):
         return f"GaussianNoise (Scale (mu={self.scale[0]}, sigma={self.scale[1]}))"
+
+
+class RotateHorizontalComponents:
+    """
+    Rotates horizontal components either by a given value alpha in rad or by an arbitrary angle between -pi and +pi.
+
+    :param alpha: Angle in rad for rotation of horizontal components.
+                  If alpha is zero, a new angle is defined each time the method is called, otherwise the angle is
```
yetinam, line 682: I think this should read "If alpha is None".

yetinam: And maybe switch it to "a random angle is selected each time ..."

seisbench/generate/augmentation.py
```diff
@@ -670,3 +672,194 @@ def __call__(self, state_dict):
...
+        self.alpha = alpha
+
+        if len(self.components) != 2:
+            msg = f"Only two components can be rotated and not {len(self.components)}."
+            raise ValueError(msg)
+
+    def __call__(self, state_dict) -> np.array:
+        x, metadata = state_dict[self.key[0]]
+
+        if self.key[0] != self.key[1]:
```
yetinam, line 714: You should also copy `x` in this case, because the operations in lines 733ff would otherwise modify the array inplace.

seisbench/generate/augmentation.py
```diff
@@ -670,3 +672,194 @@ def __call__(self, state_dict):
...
+            alpha
+        ) + data_comp2 * np.cos(alpha)
+
+        state_dict[self.key[1]] = (x, metadata)
+
+    def __str__(self):
+        if not self.alpha:
+            return f"Rotation of {self.components} components by arbitrary angle"
+        else:
+            return f"Rotation of {self.components} components by {np.round(self.alpha, 2)} rad"
```
yetinam, line 746: I'd use fstrings instead of rounding, i.e., `f"{self.alpha:.2f}"`

seisbench/generate/augmentation.py
```diff
@@ -670,3 +672,194 @@ def __call__(self, state_dict):
...
+            alpha = np.random.uniform(-np.pi, np.pi)
+        else:
+            alpha = self.alpha
+
+        trace_components = metadata.get("trace_component_order")
+        if not trace_components:
+            msg = "Keyword 'trace_component_order' is missing in metadata, therefore rotation is not possible."
+            raise ValueError(msg)
+
+        data_comp1 = copy.copy(x[trace_components.index(self.components[0]), :])
```
yetinam, line 729: I think the case that the component is not in the `trace_components` should be caught explicitly to give a more descriptive error of what's gone wrong.

seisbench/generate/augmentation.py
```diff
@@ -670,3 +672,194 @@ def __call__(self, state_dict):
...
+            return f"Rotation of {self.components} components by arbitrary angle"
+        else:
+            return f"Rotation of {self.components} components by {np.round(self.alpha, 2)} rad"
+
+
+class RealNoise:
+    """
+    Adds recorded noise from an extra noise data set to the data.
+    This method was inspired by the following publication:
+    Jun Zhu, Lihua Fang, Fajun Miao, Liping Fan, Ji Zhang, Zefeng Li
```
yetinam, line 753: Could you reformat this into a citation block? Something [like this](https://github.com/seisbench/seisbench/blob/cc180f7824b40d1fb0155e39bbf901549b709115/docs/pages/benchmark_datasets.rst?plain=1#L48).

seisbench/generate/augmentation.py
```diff
@@ -670,3 +672,194 @@ def __call__(self, state_dict):
...
+    :param key: The keys for reading from and writing to the state dict.
+                If key is a single string, the corresponding entry in state dict is modified.
+                Otherwise, a 2-tuple is expected, with the first string indicating the key to
+                read from and the second one the key to write to.
+    :type key: str, tuple[str, str]
+    :param probability: Probability of adding noise samples to data set, e.g. if probability=0.5, 50% of all samples
+                        from the dataset are deteriorated by noise. If set to 1, then each sample from the data set is
+                        deteriorated by noise. Default is 0.5
+    :type probability: float
+    :param scaling_type: Method how to find the relative amplitude of the input array. Either from the absolute maximum
```
yetinam, line 774: For consistency with the naming for the normalisation methods of the model input, could we change this to `peak` and `std`?

seisbench/generate/augmentation.py
```diff
@@ -670,3 +672,194 @@ def __call__(self, state_dict):
...
+    :param scale: Tuple of minimum and maximum relative amplitude of the noise.
+                  Relative amplitude is defined either by the absolute maximum or the root-mean-square of input array
+                  (see scaling_type). Default value is 0.5
+    :type scale: tuple(float, float)
+    :param key: The keys for reading from and writing to the state dict.
+                If key is a single string, the corresponding entry in state dict is modified.
+                Otherwise, a 2-tuple is expected, with the first string indicating the key to
+                read from and the second one the key to write to.
+    :type key: str, tuple[str, str]
+    :param probability: Probability of adding noise samples to data set, e.g. if probability=0.5, 50% of all samples
```
yetinam, line 770: I don't think this is necessary, because there is the `OneOf` augmentation. Maybe we could remove this parameter here and instead explain that only adding noise to some of the examples can be achieved using `OneOf`?

seisbench/generate/augmentation.py
```diff
@@ -670,3 +672,194 @@ def __call__(self, state_dict):
...
+                        from the dataset are deteriorated by noise. If set to 1, then each sample from the data set is
+                        deteriorated by noise. Default is 0.5
+    :type probability: float
+    :param scaling_type: Method how to find the relative amplitude of the input array. Either from the absolute maximum
+                         (max) or from the root-mean-square (rms). Default is max.
+    :type scaling_type: str
+    :param metadata_thresholds: Dictionary containing keys from metadata and threshold values as items to avoid
+                                adding noise samples, if e.g. the signal-to-noise ratio is below a certain threshold:
+                                RealNoise(noise_dataset=noise_dataset,
```
yetinam, line 779: I think this could be shortened to just `metadata_thresholds={"trace_Z_snr_db": 10}` instead of the full constructor.

seisbench/generate/augmentation.py
```diff
@@ -670,3 +672,194 @@ def __call__(self, state_dict):
...
+                    return x
+
+        # Random draw of 0 and 1 with a probability of p
+        draw = np.random.binomial(n=1, p=self.probability)
+
+        if draw == 0:
+            return x
+        else:
+            # Defining scale for noise amplitude
+            scale = np.random.uniform(*self.scale)
```
yetinam, line 836: I'm wondering if uniform is the optimal choice here. Typically, most interesting things about amplitudes are on log-scales, so we could also pick something like a truncated exponential `https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.truncexpon.html`.

seisbench/generate/augmentation.py
```diff
@@ -670,3 +672,194 @@ def __call__(self, state_dict):
...
+
+            if self.scaling_type == "max":
```
yetinam, line 838: Both of these normalisation factors will be off if the trace has a bias. I think it's necessary to remove the mean before. How about:
`np.max(np.abs(x - np.mean(x, axis=-1)))` (Also takes into account that the maximum might be negative.)
For the second normalisation, just `np.std(x)` should do the job.

seisbench/generate/augmentation.py
```diff
@@ -670,3 +672,194 @@ def __call__(self, state_dict):
...
+                scale = scale * np.max(x)
+            elif self.scaling_type == "rms":
+                scale = scale * np.sqrt(np.sum(x**2) / x.size)
+
+            # Draw random noise sample from the dataset and cut to same length as x
+            n = self.noise_generator[np.random.randint(low=0, high=self.noise_samples)][
+                self.key[0]
```
yetinam, line 845: I think you should explicitly write `"x"` here instead of `self.key[0]` because the noise generator will always write the waveform to `"x"`.

seisbench/generate/augmentation.py
```diff
@@ -670,3 +672,194 @@ def __call__(self, state_dict):
...
+            ]
+
+            # Min-max normalize noise sample and apply scale
```
yetinam, line 848: It might be more consistent to normalize with the same method as above. In addition, you should also remove the mean here. One option would be to just add an appropriately configured `Normalize` augmentation to your `noise_generator`.

seisbench/generate/augmentation.py
```diff
@@ -670,3 +672,194 @@ def __call__(self, state_dict):
...
+            n = n / np.max(np.abs(n))
+            n = n * scale
+
+            # Cutting noise to same length as x
+            spoint = np.random.randint(low=0, high=n.shape[1] - x.shape[1] - 1)
```
yetinam, line 853: What happens if the noise is not long enough? I think there should be a test case for this (and I suspect the current implementation might crash).

tests/test_generate.py
```diff
@@ -1865,6 +1867,44 @@ def test_gaussian_noise():
...
+    # Rotation by 90 degree
+    state_dict = {"X": (copy.copy(data), {"trace_component_order": "ZNE"})}
+    rotation_90 = RotateHorizontalComponents(alpha=np.pi / 2)
+    rotation_90(state_dict=state_dict)
+    assert state_dict["X"][0][0, :].all() == data[0, :].all()
+    assert state_dict["X"][0][1, :].all() == (-1 * data[2, :]).all()
+    assert state_dict["X"][0][2, :].all() == data[1, :].all()
+
+
+def test_real_noise():
```
yetinam, line 1892: This could be parametrized to test both scaling types.

tests/test_generate.py
```diff
@@ -1865,6 +1867,44 @@ def test_gaussian_noise():
     )  # Bounds are rather liberal to ensure a stable test
 
 
+def test_rotate_horizontal_components():
```
yetinam, line 1870: Can you add a random rotation? Maybe execute it twice and check that the output is different. Just to make sure this also works.

yetinam:
Hi @JanisHe ,

thanks a lot for the PR. I just went through it and it looks really good. I left a bunch of comments but I don't think any of them will require too much work. There are a few more points that didn't really fit anywhere in the code, so I just put them below.

- The parameter `key` is generally the last argument to the `__init__` function. Could you reorder it?
- The type hints are currently duplicate, once in the docstring and once in the function signature. SeisBench uses a mix of both because we started with the ones in the docstring and then switched to the signature. However, the documentation also parses the function signature, so you can remove them from the docstring. On the longer run, it would be good to fully migrate to proper type hints (#130).
- Including `N` as a marker for the noise column seems reasonable. In this context I realized that the docstring is incomplete/outdated. Could you adjust line 180 to explain how the index of the noise column is determined?

seisbench/generate/augmentation.py
JanisHe: I fully understand that point, especially to avoid circular imports.
However, at the moment, I do not have an idea how to avoid this import, since it is needed for the type hint and if I don't import the module, I get an error in line 811 for the noise generator. If you have an idea, let me know.

seisbench/generate/augmentation.py
JanisHe: I think for this case it doesn't play an important role whether the noise amplitude is uniformly or log-scale distributed. At the moment, I wouldn't change the noise scaling, but we should keep that point in mind.

JanisHe:
I @yetinam 

thank you for the suggestions. I have adopted most of your suggestions, with the exception of two (see comments). While checking the code, I found a few more errors. Hopefully the code works well now.

seisbench/generate/augmentation.py
yetinam: Okay, let's leave it at uniform for now then.

seisbench/generate/augmentation.py
yetinam: I was thinking of just dropping the type hint here altogether, but that's admittedly not a great solution. I'll update the contributing document.

seisbench/generate/augmentation.py
```diff
@@ -670,3 +672,210 @@ def __call__(self, state_dict):
...
+            raise ValueError(msg)
+
+        spoint = np.random.randint(low=0, high=n.shape[1] - x.shape[1] - 1)
+        n = n[:, spoint : spoint + x.shape[1]]
+
+        # Add noise and signal to create noisy signal
+        x = x + n
+
+        # Normalize noisy data
+        if self.scaling_type == "peak":
```
yetinam, line 870: I think it is not necessary to scale x here again and would rather leave it unscaled. The implementation above already nicely scales the noise according to x, so I think we could use the benefit of keeping x at it's original scale. Users can still put a `Normalize` module behind if they want to normalize the result. For consistency, this also would mean to remove the inplace removal of the mean on line 835 and pulling it into the calculation of the norm instead.

seisbench/generate/augmentation.py
```diff
@@ -670,3 +672,210 @@ def __call__(self, state_dict):
...
+
+        # Normalize noise samples
+        if self.scaling_type == "peak":
+            n = n / np.max(np.abs(n))
+        elif self.scaling_type == "std":
+            n = n / np.std(n)
+        n = n * scale
+
+        # Cutting noise to same length as x
+        if n.shape[1] - x.shape[1] - 1 < 0:
```
yetinam, line 856: I'm not sure I understand why there is a `-1` here. From how I understand this it should not fail for `n.shape[1] == x.shape[1]`, but it currently does.

seisbench/generate/augmentation.py
```diff
@@ -670,3 +672,210 @@ def __call__(self, state_dict):
...
+            msg = (
+                f"The shape of the data ({x.shape}) and the noise ({n.shape}) must either be the same or the "
+                f"shape of the noise must be larger than the shape of the data."
+            )
+            raise ValueError(msg)
+
+        spoint = np.random.randint(low=0, high=n.shape[1] - x.shape[1] - 1)
```
yetinam, line 863: I think this will currently fail if `n.shape[1] - x.shape[1] - 1 == 0`.

yetinam:
Thanks for all the additions! It looks almost ready now. I've made a small edit to reformat the docstrings and other minor stuff. There are only three points left that I stumble across now. Could you have a look at them? I think afterwards, it's ready to be merged.

JanisHe:
I fixed the minor issues and added a test when noise and data have the same shape

yetinam:
Thanks for the additions. I caught another small off-by-one error and fixed it and added a test. The PR is ready to merge now. Thanks a lot again for all the work and I'm looking forward to future contributions!

yetinam:
I'll also put together a v0.8 release in the next two weeks. I think there's a substantial number of new additions to warrant it.

PR #329, opened 2024-12-07
Add CWA dataset

hgvf:
Hi,

The CWA dataset is already compatible with SeisBench.
Documentation (including a dedicated figure) and the related classes have been created.
The entire dataset can be downloaded by the CWA class.

I'm excited to have this dataset as part of SeisBench, the toolbox has really helped me a lot in my research.

Thanks


yetinam:
Hi @hgvf ,
thanks again for your contribution! I finally got to look at your PR in more detail. The dataset looks really good! For the code, I'd propose a few changes:
- Currently, when downloading the dataset, you iterate over it and pass each individual trace to the writer. Given your data is already correctly formatted, this is not necessary and you just need to copy the files to the right locations. This is done, for example, in [the INSTANCE dataset](https://github.com/seisbench/seisbench/blob/e4aaba07f7a38d7c7d692abdf26421b86dbcd25e/seisbench/data/instance.py#L15). This also shows how to add the split. From the writer, you just query the `writer.metadata_path` and `writer.waveforms_path` to identify where to write the files too. Given that your dataset is chunked, you'll have to add the `chunk` argument to the signature of `_download_dataset`. Because of the chunking, it would probably be easiest if you change your tar files on hugging face to one chunk per year, as you did it for the files inside already.
- I've seen in the 2011 example that there is a `sampling_rate` set in the data format group, even though it is not constant across all examples. In this case, no sampling rate should be set. Sorry, the documentation on this is probably not too good.
- Currently, the code adds a dependency on `huggingface_hub`. This is no problem for the specific dataset, but I'd like to avoid this for SeisBench overall. The solution would be to only import the library when the dataset is actually downloaded and throw an exception telling the user to install the library if it is not available. This way, SeisBench users avoid the extra dependencies unless they want to use the CWA dataset.
- If you'd like, I can host a copy of the data on the [SeisBench servers](https://hifis-storage.desy.de:2880/Helmholtz/HelmholtzAI/SeisBench/) as well. I generally try having all datasets on there for consistency, but I understand if you don't want this. Please let me know what you think about this.
- Please make sure the code passes the automatic formatting tests. The easiest way is installing the pre-commit hooks mentioned [here](https://github.com/seisbench/seisbench/blob/main/CONTRIBUTING.md#installing-seisbench-for-development-purposes). If you run into trouble with this, I can also fix the formatting in the end before merging.
Once again, thanks for your contribution! Please don't hesitate to ask if you have any questions.

hgvf:
Hi @yetinam,
thanks for the suggestions! Sorry about the delay in getting back to you.
I've made the following revisions to the code:

- Due to the formatted data, it's no longer necessary to iterate over all the traces.
- Adding the ```chunk``` argument to the signature of ```_download_dataset```, and providing the chunk files as well.
- For the dataset, no sampling rate is set.
- The ```huggingface_hub``` library is only imported when downloading the dataset.
- Moreover, the entire CWA dataset is split into two ```BenchmarkDataset``` classes: ```CWA``` and ```CWANoise```.
- Eventually, it's welcome to maintain a copy of CWA dataset on the Seisbench server. Please feel free to use it!

p.s. I'm uploading the dataset again, it will take a few days to complete.

yetinam:
Hi @hgvf ,

thanks for this comprehensive update and sorry for taking this long to handle it. I've now went through it and I think the PR is essentially ready for merging. I did a few technical and format adjustments. Most of it is to either conform more with the style in SeisBench or to use a little more of the behind-the-scene magic. Please have a quick look if that's okay with you.

One thing I changed is that instead of downloading the split file I hard-coded the split. Given it's a simple year-wise split, it felt easier and removed an outside dependency.

The last thing I'd ask you for would be to add docstrings for the three classes. Something brief like the ones [here](https://github.com/seisbench/seisbench/blob/e4aaba07f7a38d7c7d692abdf26421b86dbcd25e/seisbench/data/instance.py#L16) is absolutely enough, given that the documentation has a longer explanation. It's just for completeness in the autogenerated part.

Before merging, I'll just do a final test downloading the whole dataset and verifying that everything works. I'll also put a copy on the SeisBench repo then.

yetinam:
I checked all the files and am uploading them to the SeisBench repository right now. I'll merge once the docstrings are added.

yetinam:
I just merged the commit. CWANoise is already in the SeisBench repo, CWA is currently uploading and should be completed in less than an hour. The dataset will be part of the v0.9 release.

Once again, thanks a lot for supporting SeisBench with your contribution!

PR #331, opened 2024-12-23
Domain-Knowledge PhaseNet  (DKPN) picker  integration

mbagagli:
Inclusion of Domain-Knowledge PhaseNet (DKPN) picker as published in Lomax et al. 2024 (https://doi.org/10.26443/seismica.v3i1.1164) and some pre-trained weights.

The code has been refactored from its original repository (https://github.com/INGV/DKPN) to suit v0.8 of seisbench (originally developed over v0.3/0.4).

Everything seems fine, but probably double checks are needed by the mantainer(s).

The end-user is referred to the original paper for details and to the original repository for the training scripts.

Additional bug fixes for `base.WaveformModel` included that double check the right device for the machine at line 489-492 during execution of the `from_pre_trained` method. This should cover cases where model-weights are trained/stored over a GPU but later used on a CPU-only machine."

### DIFF
--- a/seisbench/models/base.py
+++ b/seisbench/models/base.py
@@ -486,7 +486,10 @@ class SeisBenchModel(nn.Module):
         with open(path_json, "r") as f:
             weights_metadata = json.load(f)
         # Load model weights
-        model_weights = torch.load(f"{path_pt}")
+        #MB# : double check the right device for the machine
+        _dev = torch.device("cuda" if torch.cuda.is_available() else "cpu")
+        model_weights = torch.load(f"{path_pt}",
+                                   map_location=torch.device(_dev))
 
         model_args = weights_metadata.get("model_args", {})
         cls._check_version_requirement(weights_metadata)
--- a/seisbench/models/dkpn.py
+++ b/seisbench/models/dkpn.py
@@ -1,0 +1,784 @@
+import json
+from typing import Any
+
+import numpy as np
+import torch
+import torch.nn as nn
+import torch.nn.functional as F
+from packaging import version
+
+import seisbench.util as sbu
+from seisbench.models.base import WaveformModel, _cache_migration_v0_v3
+
+# ---------  For PreProc class
+from obspy.core import Trace
+import copy
+from obspy.signal.filter import bandpass
+from scipy.signal import lfilter
+
+
+class DKPN(WaveformModel):
+    """
+    .. document_args:: seisbench.models PhaseNet
+    """
+
+    _annotate_args = WaveformModel._annotate_args.copy()
+    _annotate_args["*_threshold"] = ("Detection threshold for the provided phase", 0.3)
+    _annotate_args["blinding"] = (
+        "Number of prediction samples to discard on each side of each window prediction",
+        (0, 0),
+    )
+    _annotate_args["overlap"] = (_annotate_args["overlap"][0], 1500)
+
+    _weight_warnings = [
+        (
+            "ethz|geofon|instance|iquique|lendb|neic|scedc|stead",
+            "1",
+            "The normalization for this weight version is incorrect and will lead to degraded performance. "
+            "Run from_pretrained with update=True once to solve this issue. "
+            "For details, see https://github.com/seisbench/seisbench/pull/188 .",
+        ),
+    ]
+
+    def __init__(
+        self,
+        in_channels=5,
+        classes=3,
+        phases="PSN",
+        sampling_rate=100,
+        norm="peak",
+        grouping="channel",
+        **kwargs,
+    ):
+        citation = (
+            "Lomax, Anthony, Matteo Bagagli, Sonja Gaviano, Spina Cianetti, "
+            "Dario Jozinović, Alberto Michelini, Christopher Zerafa, and "
+            "Carlo Giunchi. 'Effects on a Deep-Learning, Seismic Arrival-Time "
+            "Picker of Domain-Knowledge Based Preprocessing of Input "
+            "Seismograms.' Seismica 3, no. 1 (2024)."
+        )
+
+        # PickBlue options
+        for option in ("norm_amp_per_comp", "norm_detrend"):
+            if option in kwargs:
+                setattr(self, option, kwargs[option])
+                del kwargs[option]
+            else:
+                setattr(self, option, False)
+
+        super().__init__(
+            citation=citation,
+            in_samples=3001,
+            output_type="array",
+            pred_sample=(0, 3001),
+            labels=phases,
+            sampling_rate=sampling_rate,
+            default_args={
+                "overlap": 1500,
+                "P_threshold": 0.2,
+                "S_threshold": 0.2,
+                #
+                "fp_stabilization": 4,  # sec
+                "t_long": 4,
+                "freqmin": 0.5,
+                "corner": 1,
+                "perc_taper": 0.1,
+                "mode": "rms",
+                "clip": -999,
+                "log": True,
+                "normalize": False,
+                "polarization_win_len": 1,
+                "use_amax_only": False},
+            **kwargs,
+        )
+
+        self.in_channels = in_channels
+        self.classes = classes
+        self.norm = norm
+        self.depth = 5
+        self.kernel_size = 7
+        self.stride = 4
+        self.filters_root = 8
+        self.activation = torch.relu
+
+        self.inc = nn.Conv1d(
+            self.in_channels, self.filters_root, self.kernel_size, padding="same"
+        )
+        self.in_bn = nn.BatchNorm1d(8, eps=1e-3)
+
+        self.down_branch = nn.ModuleList()
+        self.up_branch = nn.ModuleList()
+
+        last_filters = self.filters_root
+        for i in range(self.depth):
+            filters = int(2**i * self.filters_root)
+            conv_same = nn.Conv1d(
+                last_filters, filters, self.kernel_size, padding="same", bias=False
+            )
+            last_filters = filters
+            bn1 = nn.BatchNorm1d(filters, eps=1e-3)
+            if i == self.depth - 1:
+                conv_down = None
+                bn2 = None
+            else:
+                if i in [1, 2, 3]:
+                    padding = 0  # Pad manually
+                else:
+                    padding = self.kernel_size // 2
+                conv_down = nn.Conv1d(
+                    filters,
+                    filters,
+                    self.kernel_size,
+                    self.stride,
+                    padding=padding,
+                    bias=False,
+                )
+                bn2 = nn.BatchNorm1d(filters, eps=1e-3)
+
+            self.down_branch.append(nn.ModuleList([conv_same, bn1, conv_down, bn2]))
+
+        for i in range(self.depth - 1):
+            filters = int(2 ** (3 - i) * self.filters_root)
+            conv_up = nn.ConvTranspose1d(
+                last_filters, filters, self.kernel_size, self.stride, bias=False
+            )
+            last_filters = filters
+            bn1 = nn.BatchNorm1d(filters, eps=1e-3)
+            conv_same = nn.Conv1d(
+                2 * filters, filters, self.kernel_size, padding="same", bias=False
+            )
+            bn2 = nn.BatchNorm1d(filters, eps=1e-3)
+
+            self.up_branch.append(nn.ModuleList([conv_up, bn1, conv_same, bn2]))
+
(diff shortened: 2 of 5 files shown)


yetinam:
Hi @mbagagli ,
thanks a lot for submitting this PR! I'll have a look at it in the new year. Happy holidays!

seisbench/models/base.py
```diff
@@ -486,7 +486,10 @@ def load(cls, path, version_str=None):
         with open(path_json, "r") as f:
             weights_metadata = json.load(f)
         # Load model weights
-        model_weights = torch.load(f"{path_pt}")
+        #MB# : double check the right device for the machine
+        _dev = torch.device("cuda" if torch.cuda.is_available() else "cpu")
```
yetinam, line 490: Thanks for the fix. I'd suggest a small change, in that I think by default we should load models always to CPU. Users can still move them to GPU manually afterwards. As our models are small, this doesn't have any performance downsides and it will keep the GPU usage explicit.

seisbench/models/dkpn.py
```diff
@@ -0,0 +1,784 @@
...
+from seisbench.models.base import WaveformModel, _cache_migration_v0_v3
+
+# ---------  For PreProc class
+from obspy.core import Trace
+import copy
+from obspy.signal.filter import bandpass
+from scipy.signal import lfilter
+
+
+class DKPN(WaveformModel):
```
yetinam, line 20: Given this is essentially a PhaseNet model, it should inherit from PhaseNet

seisbench/models/dkpn.py
```diff
@@ -0,0 +1,784 @@
...
+    """
+    .. document_args:: seisbench.models PhaseNet
```
yetinam, line 22: This should read `.. document_args:: seisbench.models DKPN`

seisbench/models/dkpn.py
```diff
@@ -0,0 +1,784 @@
...
+    """
+
+    _annotate_args = WaveformModel._annotate_args.copy()
```
yetinam, line 25: `default_args` should be specified (and documented) here, so that they appear correctly in the documentation.

seisbench/models/dkpn.py
```diff
@@ -0,0 +1,784 @@
...
+                2 * filters, filters, self.kernel_size, padding="same", bias=False
+            )
+            bn2 = nn.BatchNorm1d(filters, eps=1e-3)
+
+            self.up_branch.append(nn.ModuleList([conv_up, bn1, conv_same, bn2]))
+
+        self.out = nn.Conv1d(last_filters, self.classes, 1, padding="same")
+        self.softmax = torch.nn.Softmax(dim=1)
+
+        # DKPN attributes for taking care of windows CF and probs
```
yetinam, line 157: Any transient attributes should not be stored in class variables. From what I see, these variable are actually only written to but never read. Can they be deleted?

seisbench/models/dkpn.py
```diff
@@ -0,0 +1,784 @@
...
+        self.batch_ = []
+        self.batch_cfs = []
+        self.batch_probs = []
+        self.stream_cfs = None
+
+    def __reset_predict(self):
```
yetinam, line 163: See comment above

seisbench/models/dkpn.py
```diff
@@ -0,0 +1,784 @@
...
+    # ==================================  PROCESS STREAM
+    # ================================================================
+
+    def annotate_stream_pre(self, stream, argdict):
+        """ ARGDICT represent the default dict,merged (eventually)
+            with the key:value pairs of the "*.json" specifying the model
+        """
+        self.__reset_predict()
+
+        # -----------------   Pre-pend the original code of SEISBENCH (v0.2)
```
yetinam, line 219: I think you can just call `stream = super().annotate_stream_pre(stream, argdict)` instead.

seisbench/models/dkpn.py
```diff
@@ -0,0 +1,784 @@
...
+            "blinding", self._annotate_args.get("blinding")[1]
+        )
+        if prenan > 0:
+            batch[:, :prenan] = np.nan
+        if postnan > 0:
+            batch[:, -postnan:] = np.nan
+        self.batch_probs.append(batch.detach().clone())
+        return batch
+
+    def extract_cf(self, stream):
```
yetinam, line 275: The function and the two below are never called.

seisbench/models/dkpn.py
```diff
@@ -0,0 +1,784 @@
...
+            del model_args[key]
+
+        model_args["in_channels"] = self.in_channels
+        model_args["classes"] = self.classes
+        model_args["phases"] = self.labels
+        model_args["sampling_rate"] = self.sampling_rate
+
+        return model_args
+
+    def get_defaults(self):
```
yetinam, line 340: Given `default_args` is a public variable, I think it's unnecessary adding a wrapper function.

seisbench/models/dkpn.py
```diff
@@ -0,0 +1,784 @@
...
+        return self.default_args
+
+    def set_dkpn_parameter(self, pardict):
```
yetinam, line 343: ... or a setter function for this mapper.

seisbench/models/dkpn.py
```diff
@@ -0,0 +1,784 @@
...
+
+    _annotate_args = WaveformModel._annotate_args.copy()
+    _annotate_args["*_threshold"] = ("Detection threshold for the provided phase", 0.3)
+    _annotate_args["blinding"] = (
+        "Number of prediction samples to discard on each side of each window prediction",
+        (0, 0),
+    )
+    _annotate_args["overlap"] = (_annotate_args["overlap"][0], 1500)
+
+    _weight_warnings = [
```
yetinam, line 33: This is unnecessary here. It warns about outdated weights versions but for DKNP we do not have these.

seisbench/models/dkpn.py
```diff
@@ -0,0 +1,784 @@
...
+# ====================================================================
+# ====================================================================
+# ====================================================================
+# ====================================================================
+
+
+class PreProc(object):
+    def __init__(self, **kwargs):
+        self.stream = None  # --> Instantiated everytime in the work() class-method
+        self.__dict__.update(kwargs)
```
yetinam, line 436: This works, but could you explicitly define the arguments in the function header and safe them explicitly. This will make documentation and debugging much easier.

seisbench/models/dkpn.py
```diff
@@ -0,0 +1,784 @@
...
+# ====================================================================
+# ====================================================================
+# ====================================================================
+# ====================================================================
+# ====================================================================
+
+
+class PreProc(object):
+    def __init__(self, **kwargs):
+        self.stream = None  # --> Instantiated everytime in the work() class-method
```
yetinam, line 435: As for the DKPN class, temporary variables should not be saved in class variables. Instead, you could make `work` return the processed stream.

seisbench/models/dkpn.py
```diff
@@ -0,0 +1,784 @@
...
+        Provides an augmentation for seisbench.generate.generator.GenericGenerator.add_augmentations().
+        """
+
+        labels, metadata_labels = state_dict["y"]
+        waveforms, metadata_waveforms = state_dict["X"]  # metadata is the relevant row-field of CSV
+        waveforms = self.__matrix_cfs__(waveforms)
+
+        # ================================  Emulates WINDOW-LEVEL NORM
+        # --- Remove FP stabilization
+        fstab = int(metadata_waveforms["trace_sampling_rate_hz"]*self.fp_stabilization)
```
yetinam, line 450: I'm not sure I understand these lines. I guess they should remove some padding in the beginning, but they also truncate at the end?

seisbench/models/dkpn.py
```diff
@@ -0,0 +1,784 @@
...
+
+        # --- NORMALIZE  MODULUS matrix with the its STD
+        waveforms[4, :] = waveforms[4, :] / (
+            np.std(waveforms[4, :], axis=-1, keepdims=True) + 1e-10
+            )
+
+        state_dict["X"] = (waveforms, metadata_waveforms)  # OVERRIDE MATRIX --> OUTPUT
+        state_dict["y"] = (labels, metadata_labels)  # OVERRIDE MATRIX --> OUTPUT
+
+    def __matrix_cfs__(self, MM):
```
yetinam, line 467: I think this should be `_matrix_cfs` instead as private functions in python typically start with a single underscore and don't use trailing underscores.

seisbench/models/dkpn.py
```diff
@@ -0,0 +1,784 @@
...
+        fp_band_data = []
+        fp_max_arrays_band = []
+        fp_max_arrays_val = []
+
+        for _idx, arr in enumerate(MM):
+
+            # ====================
+            # 1) Run FBsummary
+            summary = FBSummary(arr,
+                                sampling_rate=100.0,
```
yetinam, line 489: Is it intended that the sampling rate is hard-coded here?

seisbench/models/dkpn.py
```diff
@@ -0,0 +1,784 @@
...
+
+        # --- Bring everything to float32
+        fp_cf_waveforms = fp_cf_waveforms.astype("float32")
+
+        return fp_cf_waveforms
+
+    def get_stream(self):
+        return self.stream
+
+    def work(self, stream):
```
yetinam, line 547: From what I can see, this function assumes that the stream consists of exactly three traces, named `ZNE`, from a single station, and aligned in time. This is not given for streams passed to `annotate`. Instead, this work should be done in `annotate_batch_pre`.

seisbench/models/dkpn.py
```diff
@@ -0,0 +1,784 @@
...
+                self.stream += Trace(data=_cfdata,
+                                     header=_cfstats)
+
+        # # ==================== FP_ STABILIZATION
+        # print("DEBUG: removing the first %.2f seconds of trace" % self.fp_stabilization)
+        # for _tr in self.stream:
+        #     _tr.trim(_tr.stats.starttime + self.fp_stabilization,
+        #              _tr.stats.endtime)
+
+    def PolarizationFP(self, fp_band_data, fp_max_arrays_band_amax,
```
yetinam, line 602: Function names should be snake case and start with a non-capital letter.

seisbench/models/dkpn.py
```diff
@@ -0,0 +1,784 @@
...
+        for j in range(n_bands):
+            octave_high = (self.freqmin + self.freqmin * 2.0) / 2.0 * (2**j)
+            octave_low = octave_high / 2.0
+            BF[j] = bandpass(self.data, octave_low, octave_high,
+                             self.sampling_rate,
+                             corners=self.cnr,
+                             zerophase=False)
+        return BF
+
+    def get_summary(self):
```
yetinam, line 741: I think the wrapper can be deleter.

yetinam:
Hi @mbagagli ,
now I got to look through your PR. Thanks again for the contribution! There are a few points that need to be adjusted:

- The class copies a lot of code from PhaseNet. Following the DRY principle, instead the class should inherit from PhaseNet and just overwrite the necessary parts. This will also make reviewing easier.
- The `annotate` function does not work (at least for me). I've put a comment in the code where it crashes. The problem is that your modified inputs aren't correctly passed in. I think, you should do the following: (1) in `annotate_stream_pre` calculate the FilterPicker traces. From what I understood this is done per component, so it's pretty straightforward. Note that at this stage SeisBench has not aligned the different traces. Write the outputs into the stream with components like `FPA`, `FPB`, `FPC`. (2) Set the `component_order` to `ZNEABC`. This will force the model to cut give 6 channel data to `annotate_batch_pre`. (3) In `annotate_batch_pre`, calculate modulus and inclination. At this stage, SeisBench has aligned the traces, split them into different stations, ... . This means, the calculations are simple matrix operations. Afterwards, discard the channels you don't actually want to pass to the model (ZNE I guess). With these three steps, everything should work.
- Currently, the implementation saved temporary data into class variables, making the class stateful. Unless absolutely necessary, this should be avoided.
- To keep the size of the repository smaller, no weights are stored directly in the repository. However, I'll upload them to the SeisBench repository to make them generally available.
- Please add a unit test for DKPN. They can be very simple (see the examples already implemented), just checking that if you pass a valid input stream, the output of  `annotate` is not empty.
- Please run the [precommit hooks](https://github.com/seisbench/seisbench/blob/main/CONTRIBUTING.md#submitting-a-pull-request) to adjust the formatting.

Let me know if you have any questions and I'll be happy to help.

PR #336, opened 2025-03-12
Add the CEED dataset to SeisBench

yetinam:
This PR adds the CEED dataset from Zhu et al. (2025) to SeisBench. Details on the dataset are available [here](https://huggingface.co/datasets/AI4EPS/CEED) and [here](https://arxiv.org/pdf/2502.11500).

@zhuwq0 I hope the addition is okay with you. There are a few technical questions I wanted to check with you:
- I'd like to include a figure in the documentation, showing the dataset. Figure 1 from your preprint would be perfect for that. Would it be possible to use that?
- Would it be okay if I host a copy of CEED in SeisBench format in the SeisBench repo? I'm trying to keep all dataset in this repo, because it's the quickest way for users to access the data and does not require conversion. As you can see, the conversion from the original files to the SeisBench format is pretty straight forward for CEED, but it still takes a while to compile the metadata.

**NOTE:** Only merge once the TODO notes in the code have been addressed.


zhuwq0:
Thank you for your work on integrating the CEED dataset into SeisBench. I really appreciate your efforts to enhance accessibility for users.
Please feel free to use any figures and host a copy of the CEED dataset in SeisBench format.

PR #355, opened 2025-05-23
skynet integration

albertleonardo:
Add regional and multiphase pickers from skynet


yetinam:
I just merged the PR. Skynet is now available in main and will be part of the v0.10 release. Thanks again for adding this so quickly!

PR #360, opened 2025-06-25
Training an own Denoiser model

JanisHe:
This pull request contains classes, functions and a Jupyter-Notebook to train a DeepDenoiser / SeisDAE model.

- SeisDAE: A fully trainable CNN to train a denoiser model using the short-time Fourier transform (STFT). The normalization of the input data differs in comparison to the original DeepDenoiser, i.e. in SeisDAE the waveform data are min-max normalized and in a second step the real and imaginary parts of the STFT are also min-max normalized and DeepDenoiser only normalizes the the real and imaginary parts of the STFT by standardization.
The depth, kernel size, stride, droprate, filters root, nfft and nperseg can be easily modified by the user. Also attention gates can be used, however, I did not notice any improvement but perhaps somebody else does.
- STFTDenoiserLabeller: A labeling class that creates the training data by taking high SNR earthquake waveforms and noise waveforms, adding both, computing STFT for earthquake, noise and noisy waveform and target mask functions.  
- Training Denoiser Notebook: The notebook gives an introduction how to train an own SeisDAE model and also includes easy code that describes how to build a noise dataset. However, building the noise dataset can be done more accuratly by users, which is mentioned in the notebook.


yetinam:
Hi @JanisHe ,
thanks a lot for the big addition. I think this is going to be very useful. I'll look through it soon. A few small things I noticed so far:
- Could you add the new tutorial also to the README in the tutorials folder and to the relevant documentation page?
- Could you add the new denoiser to the models documentation page?
- In addition, I'd ask you to verify whether the class is already documented automatically as it's in deepdenoiser.py or if it needs to be added explicitly.

JanisHe:
Hi @yetinam,

I've updated the Readmes and documentation, however, I don't think the new SeisDAE class is documented automatically, since I don't know how to do this. Perhaps you might help here.

One point that is not solved well in the labeling class: Adding noise to the earthquake waveform data uses the same code as for the RealNoise class in augmentations. I didn't find a solution how to implement this in a better Pythonic way....

seisbench/generate/labeling.py
```diff
@@ -774,6 +776,208 @@ def __str__(self):
...
+                Default key is ("X", "y").
+    :param nfft: Length of the FFT used, if a zero padded FFT is desired for scipy STFT.
+                 If None, the FFT length is nperseg.
+    :param nperseg: Length of each segment for scipy STFT. Default is 60
+    :param eps: Avoiding division by zero, default is 1e-12.
+    """
+
+    def __init__(
+        self,
+        noise_dataset: seisbench.data.base.WaveformDataset,
```
yetinam, line 812: I'd use `seisbench.data.WaveformDataset` to avoid the internal API.

seisbench/generate/labeling.py
```diff
@@ -774,6 +776,208 @@ def __str__(self):
...
+        scale: tuple[float, float] = (0, 1),
+        scaling_type: str = "peak",
+        component: str = "ZNE",
+        sampling_rate: Union[float, None] = None,
```
yetinam, line 816: `Optional[float]` makes the type hint more compact.

yetinam, line 816: `Optional[float]` makes the type hint more compact.

seisbench/generate/labeling.py
```diff
@@ -774,6 +776,208 @@ def __str__(self):
...
+            metadata["trace_component_order"] = self.component[component_idx]
+
+        # Select component from x and modify metadata
+        x = x[component_idx, :]
+
+        # Defining scale for earthquake and noise amplitude
+        # I.e., scale for earthquake and noise can become, for example, zero and thus either no noise or
+        # earthquake waveform exists. Using this approach, the denoiser learns better to distinguish between noise
+        # and earthquake.
+        scale_earthquake = np.random.uniform(*self.scale)
```
yetinam, line 866: I'm not sure this is consistent with the docstring. From the docstring I assumed that the scale describes the relative amplitude of the noise compared to the signal. However, here both are scaled. Maybe this should be described explicitly in the docstring?

seisbench/generate/labeling.py
```diff
@@ -774,6 +776,208 @@ def __str__(self):
...
+        # Normalize earthquake and noise samples
+        if self.scaling_type == "peak":
+            x = x / np.max(np.abs(x) + self.eps)
+            n = n / (np.max(np.abs(n)) + self.eps)
+        elif self.scaling_type == "std":
+            x = x / (np.std(x) + self.eps)
+            n = n / (np.std(n) + self.eps)
+
+        # Cutting noise to same length as x
+        if len(n) - len(x) < 0:
```
yetinam, line 888: I'd suggest `n.shape[-1] < x.shape[-1]`

seisbench/generate/labeling.py
```diff
@@ -774,6 +776,208 @@ def __str__(self):
...
+            msg = (
+                f"The length of the data ({len(x)} samples) and the noise ({len(n)} samples) must either be the same "
+                f"or the shape of the noise must be larger than the shape of the data."
+            )
+            raise ValueError(msg)
+
+        if len(n) - len(x) > 0:
```
yetinam, line 895: Same as above

yetinam, line 895: Same as above

seisbench/generate/labeling.py
```diff
@@ -774,6 +776,208 @@ def __str__(self):
...
+        n = n[component_idx, :]
+
+        # Removing mean from earthquake and noise samples
+        x = x - np.mean(x, axis=-1, keepdims=True)
+        n = n - np.mean(n, axis=-1, keepdims=True)
+
+        # Normalize earthquake and noise samples
+        if self.scaling_type == "peak":
+            x = x / np.max(np.abs(x) + self.eps)
+            n = n / (np.max(np.abs(n)) + self.eps)
```
yetinam, line 882: Total nitpick, but why are the brackets in lines 881 and 882 different?

seisbench/generate/labeling.py
```diff
@@ -774,6 +776,208 @@ def __str__(self):
...
+            shape=(2, *stft_n.shape)
+        )  # Input, i.e. real and imag part of noisy
+
+        y = np.empty(
+            shape=(2, *stft_n.shape)
+        )  # Target, i.e. masks for signal and noise
+
+        # Normalize real and imaginary parts of STFT to range [-1, 1]
+        X[0, :, :] = stft_noisy.real / (np.max(np.abs(stft_noisy.real)) + self.eps)
+        X[1, :, :] = stft_noisy.imag / (np.max(np.abs(stft_noisy.imag)) + self.eps)
```
yetinam, line 953: I think normalizing real and imaginary part independently causes issues, because it will distort both phase and amplitude of the complex signal.

seisbench/generate/labeling.py
```diff
@@ -774,6 +776,208 @@ def __str__(self):
...
+        # Compute target masking function
+        y[0, :, :] = 1 / (1 + np.abs(stft_n) / (np.abs(stft_x) + self.eps))
+        y[1, :, :] = 1 - y[0, :, :]
+
+        # Replace nan values in X and y
+        X = np.nan_to_num(X)
+        y = np.nan_to_num(y)
+
+        # Update state_dicts for input and target
+        state_dict[self.key[0]] = (X, metadata)
```
yetinam, line 964: `self.key[0]` is the input key and should not be overwritten (unless explicitly requested by the user). Maybe you could make the parameters to the class more specific. Something like `input_key="X", noisy_output_key="X", mask_key="y"`. Then it would be more explicit that the module will overwrite `"X"`.

seisbench/generate/labeling.py
```diff
@@ -774,6 +776,208 @@ def __str__(self):
...
+        state_dict[self.key[1]] = (y, metadata)
+
+    def label(self, X, metadata):
```
yetinam, line 967: I'm wondering if it might be an option to not subclass `SupervisedLabeller`. Semantically it seems reasonable, but on the other hand I think you're not actually using any functions of the superclass.

seisbench/models/deepdenoiser.py
```diff
@@ -284,3 +283,454 @@ def forward(self, x, mid):
...
+            "Journal of Seismology, 26(5), 913-934. "
+            "https://doi.org/10.1007/s10950-022-10097-6"
+        )
+
+        WaveformModel.__init__(
+            self,
+            citation=citation,
+            in_samples=in_samples,
+            output_type="array",
+            default_args={"overlap": in_samples // 2},
```
yetinam, line 490: Setting this will make it impossible to pass `default_args` through the kwargs, so it should be removed. It is also not necessary, because it is set before (see comment above).

seisbench/models/deepdenoiser.py
```diff
@@ -284,3 +283,454 @@ def forward(self, x, mid):
...
+    :param nfft: Length of the FFT used, if a zero padded FFT is desired for scipy STFT.
+                 If None, the FFT length is nperseg.
+    :param nperseg: Length of each segment for scipy STFT. Default is 60
+    :param attention: Whether to use attention gates in skip connections or not.
+                      Default is False
+    :param kwargs: Additional arguments passed to the base `WaveformModel`.
+    """
+
+    _annotate_args = WaveformModel._annotate_args.copy()
+    _annotate_args["overlap"] = (_annotate_args["overlap"][0], 1500)
```
yetinam, line 452: You can change it from `1500` to `0.5` to automatically use half the window length as overlap. I added this option recently.

seisbench/models/deepdenoiser.py
```diff
@@ -284,3 +283,454 @@ def forward(self, x, mid):
...
+        return x
+
+    def annotate_batch_pre(
+        self, batch: torch.Tensor, argdict: dict[str, Any]
+    ) -> torch.Tensor:
+        """
+        Does preprocessing for prediction
+        """
+        # STFT of each batch and component
+        self.norm_factors = np.empty(shape=batch.shape[0])
```
yetinam, line 659: I think it would be better to avoid writing to a class variable here. Just add `norm_factors` to the piggyback by returning `noisy_input, (noisy_stft, norm_factors)` at the bottom.

seisbench/models/deepdenoiser.py
```diff
@@ -284,3 +283,454 @@ def forward(self, x, mid):
...
+        noisy_input[torch.isnan(noisy_input)] = 0
+        noisy_input[torch.isinf(noisy_input)] = 0
+
+        return noisy_input, noisy_stft
+
+    def _normalize_trace(self, batch: torch.Tensor):
+        """
+        Normalize each trace and save norm factors to scale back denoised traces.
+        """
+        for trace_id in range(batch.shape[0]):
```
yetinam, line 686: This should be vectorized for performance.

seisbench/models/deepdenoiser.py
```diff
@@ -284,3 +283,454 @@ def forward(self, x, mid):
...
+        return batch
+
+    def annotate_batch_post(
+        self, batch: torch.Tensor, piggyback: Any, argdict: dict[str, Any]
+    ) -> torch.Tensor:
+        """
+        Does postprocessing when predicting datasets
+        """
+        # Multiply piggyback (STFT of noisy signal) with batch (predicted mask) for signal
+        signal_stft = piggyback * batch.numpy()[:, 0, :]
```
yetinam, line 706: Same as above, the `annotate_batch_post` function should run in torch space.

seisbench/util/torch_helpers.py
```diff
@@ -18,3 +18,26 @@ def worker_seeding(wid):
...
+    :param eps: Float to avoid division by zeros
+    :return: min-max normalized input array
+    """
+    return (x - np.min(x)) / (np.max(x) - np.min(x) + eps)
+
+
+def z_score_normalization(x: np.array) -> np.array:
+    """
+    Normalize data by z-score
+    .. math::
```
yetinam, line 37: The math doesn't render correctly. I think it needs an additional blank line before the `.. math::` statement.

yetinam, line 37: The math doesn't render correctly. I think it needs an additional blank line before the `.. math::` statement.

tests/test_models.py
```diff
@@ -1098,6 +1098,16 @@ def test_annotate_deep_denoiser():
         assert annotations[i].data.shape == (3000,)
 
 
+def test_annotate_seisdae():
+    stream = obspy.read()
```
yetinam, line 1102: Maybe it would be better to replace this with an artificially created random stream of fixed length. Then the function could also be parametrized to check that it also works if the input doesn't exactly have the length of the window size.

seisbench/generate/labeling.py
JanisHe: I forgot to add the brackets in the first statement, otherwise it doesn't make sense, when trying to avoid division by zero.

yetinam: Should be fine either way, given the `+ self.eps` is outside the `np.abs` call in both cases.

yetinam:
Hi @JanisHe,

thanks for the update. Just for the overview, can you mark my comments as resolved once fixed? I'll have another look at the PR then.

For the random module, what is the problem with exchanging it? Usually I'd assume any functionality of `random` can also be achieved with `np.random`, but I might be wrong. However, if it's just for the notebook I also don't have particularly strong feelings about it.

yetinam:
And thanks for the update about individual normalization of real and imaginary part. That's interesting! Let's keep it this way then. I'd be curious to figure out at some point why individually normalizing the two parts works better.

JanisHe:
> Hi @JanisHe,
> 
> thanks for the update. Just for the overview, can you mark my comments as resolved once fixed? I'll have another look at the PR then.
> 
> For the random module, what is the problem with exchanging it? Usually I'd assume any functionality of `random` can also be achieved with `np.random`, but I might be wrong. However, if it's just for the notebook I also don't have particularly strong feelings about it.

I have no idea why it did not work, since both methods take a list. However, when I use numpy I always end up with the following error: ValueError: setting an array element with a sequence. The requested array has an inhomogeneous shape after 1 dimensions. The detected shape was (5,) + inhomogeneous part.

I don't know why this happens, however, I think for the notebook it doesn't play an important role.

seisbench/generate/labeling.py
```diff
@@ -774,6 +776,208 @@ def __str__(self):
...
+                  Relative amplitude is defined either by the absolute maximum or the root-mean-square of input array
+                  (see scaling_type). Default value is (0, 1).
+    :param scaling_type: Method how to find the relative amplitude of the input array. Either from the absolute maximum
+                         (peak) or from the standard deviation (std). Default is peak.
+    :param component: Component to take into account for training. Default are three components, ZNE.
+                      If DeepDenoiser/SeisDAE is only trained with a single component, use e.g., component="Z".
+                      Otherwise, DeepDenoiser/SeisDAE is trained with all components, however, noise samples are only
+                      added on same component of earthquake waveforms, i.e., on Z-component are only added noise
+                      waveforms of Z-component and so on.
+    :param sampling_rate: Sampling rate in Hz of earthquake waveform and noise dataset.
```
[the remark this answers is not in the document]
JanisHe: I will figure out whether I can delete the parameter sampling rate. Thanks for that hint.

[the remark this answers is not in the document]
JanisHe: This parameter is not necessary anymore.

seisbench/generate/labeling.py
```diff
@@ -774,6 +775,203 @@ def __str__(self):
...
+        self.kwargs = kwargs
+
+        if self.scaling_type.lower() not in ["std", "peak"]:
+            msg = "Argument scaling_type must be either 'std' or 'peak'."
+            raise ValueError(msg)
+
+    def __call__(self, state_dict):
+        x, metadata = state_dict[self.input_key]
+
+        if self.input_key != self.mask_key:
```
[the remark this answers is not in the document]
JanisHe: Yeah, you are totally right. I resolved this point. Thanks :)

yetinam:
I made a few (very small) edits to the notebook and uploaded the weights [to the repo](https://hifis-storage.desy.de:2880/Helmholtz/HelmholtzAI/SeisBench/models/v3/seisdae/). I'll wait for the CI to pass and merge then. Thanks a lot again for this contribution and for addressing all my comments swiftly!

PR #368, opened 2025-07-18
pre-commit: using ruff

miili:
Replaces black and isort.


yetinam:
Looks good. I'm a little afraid of the rebasing for the existing PRs though because the change in the precommit hooks touches most of the codebase including the jupyter notebooks. Is there a smart way for that? Otherwise, I'd suggest to wait until the other PRs are merged, than rebase this PR (which should be pretty straightforward) and then merge.

PR #372, opened 2025-08-13
[bugfix] Labelling for denoised channels was not set

JanisHe:
When applying the Denoiser on a stream, each trace had the same label. This is fixed with this PR.


yetinam:
Thanks for catching and fixing this. I've cherry-picked it to the maintenance branch as well and pushed it to PyPI as v0.10.1.

PR #373, opened 2025-08-20
Add PhaseNetPlus

### LINKED ISSUE #91: Add polarity picking model(s)
It would be great to add polarity picking models to SeisBench. This should be relatively easy to do and we already have training data. At the same time, it would be very useful to derive focal mechanisms.

I'm not fully up to date on the literature but there is at least the paper by Ross et al.

yetinam:
Add the PhaseNetPlus model. The weight set "original" has been uploaded to the [SeisBench repository](https://hifis-storage.desy.de:2880/Helmholtz/HelmholtzAI/SeisBench/models/v3/phasenetplus/).

Please have a look @wanghaoyucn and @zhuwq0

### DIFF
--- a/docs/pages/documentation/models.rst
+++ b/docs/pages/documentation/models.rst
@@ -100,7 +100,22 @@ OBSTransformer
 PhaseNet
 -----------------------------------
 
-.. automodule:: seisbench.models.phasenet
+.. autoclass:: seisbench.models.phasenet.PhaseNet
+    :members:
+    :undoc-members:
+    :show-inheritance:
+
+.. autoclass:: seisbench.models.phasenet.PhaseNetLight
+    :members:
+    :undoc-members:
+    :show-inheritance:
+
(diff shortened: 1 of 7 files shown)


yetinam:
Thanks for the quick reply! I've added the tolerance for the polarity extraction. For the pick association, there is already a tolerance built-in. The implementation looks in the segment `[time_est - pick_tolerance, time_est + pick_tolerance]` for picks (see code below). `time_est` is estimated from the event center and the event time, `pick_tolerance` is set in the constructor, by default to 3 seconds. Do you think this tolerance is big enough?

https://github.com/seisbench/seisbench/pull/373/files#diff-8cf308ac287f0e1b5585eabd15851c066e192f39365be35928f949afe979da80R976-R980

miili:
@yetinam, what is blocking this feature?

miili:
@yetinam is there an update on the publication status?

PR #376, opened 2025-09-18
Dataset: Adding pyrocko inspection methods

miili:
Add inspection methods to `WaveformDataset` to snuffle individual events as well as dump the whole dataset to MiniSeed alongside all event, picking, and station information.

image


seisbench/data/base.py
```diff
@@ -2677,3 +2706,53 @@ def flush_hdf5(self):
         for bucket in buckets:
             self._write_bucket(self._cache[bucket])
             del self._cache[bucket]
+
+
+class EventParameters(TypedDict):
+    index: NotRequired[int]
+    split: Literal["train", "dev", "test"]
+
+    source_id: str
```
yetinam: Can you label all of these parameters as `NotRequired`. The data format is supposed to be very permissive.

seisbench/data/inspection.py
```diff
@@ -0,0 +1,410 @@
...
+        :return: List of pyrocko Trace objects
+        """
+
+        ds = self._dataset
+
+        data, metadata = ds.get_sample(sample_idx)
+        traces = []
+
+        location_code = metadata.get("station_location_code")
+        location_code = "" if np.isnan(location_code) else location_code
```
yetinam: This will fail if the location code is not numerical.

yetinam:
Hey @miili ,

sorry for the long delay, but now I got to looking at the PR. I've added a few small comments but overall the PR looks good to me. It'll be a very useful addition to improve the exploration of the dataset. We might consider adding a few lines somewhere in the documentation or examples pointing out the possibility of inspecting datasets like this.

I haven't tried our the implementation yet, but I'll give it a go tomorrow and let you know if there's anything else.

seisbench/data/inspection.py
pevans-gfz: Indeed, location codes in SEED are strings. 0-2 alphanumeric uppercase (or dash "-"!) characters today, up to eight characters in future. Note "01" != "1", "00" != "". https://docs.fdsn.org/projects/source-identifiers/en/latest/definition.html

seisbench/data/inspection.py
```diff
@@ -0,0 +1,410 @@
...
+        all_traces, event, all_picks, stations = self._get_pyrocko_data(event)
+
+        snuffle(
+            traces=all_traces,
+            events=[event],
+            markers=all_picks,
+            stations=[sta.as_pyrocko_station() for sta in stations],
+        )
+
+    def export_mseed(self, directory: Path) -> None:
```
yetinam: For convenience I'd suggest allowing people to use string arguments as well.

seisbench/data/inspection.py
```diff
@@ -0,0 +1,410 @@
...
+- `picks/`: Contains pick files in pyrocko .picks format, one file per day.
+- `events.yaml`: Contains all events in pyrocko .yaml format.
+- `stations.yaml`: Contains all stations in pyrocko .yaml format.
+- `all_picks.picks`: Contains all picks in pyrocko .picks format
+
+## Open Pyrocko Snuffler
+
+You can open the data in pyrocko's snuffler using the following command:
+```bash
+squirrel snuffler -a mseed/*.mseed -e events.yaml -s stations.yaml
```
yetinam: Do the picks need to be loaded explicitly as well or is that part of loading the events?

PR #380, opened 2025-10-01
Add EQTP model

Pengling2022:
EQTP is a multi-task model for seismic phase picking and P-wave first motion polarity determination, building upon the original version of the EQTransformer model.

### DIFF
--- a/seisbench/models/eqtp.py
+++ b/seisbench/models/eqtp.py
@@ -6,7 +6,16 @@ import torch
 import torch.nn as nn
 
 import seisbench.util as sbu
-from .eqtransformer import EQTransformer, SeqSelfAttention, Decoder, Encoder, ResCNNStack, Transformer, ActivationLSTMCell, CustomLSTM
+from .eqtransformer import (
+    EQTransformer,
+    SeqSelfAttention,
+    Decoder,
+    Encoder,
+    ResCNNStack,
+    Transformer,
+    ActivationLSTMCell,
+    CustomLSTM,
+)
 
 
 class EQTP(EQTransformer):
@@ -34,13 +43,8 @@ class EQTP(EQTransformer):
     def __init__(
         self,
         in_samples=12000,
-        cnn_blocks=5,
-        res_cnn_blocks=5,
         **kwargs,
     ):
-        self.cnn_blocks = cnn_blocks
-        self.res_cnn_blocks = res_cnn_blocks
-
         # Update citation for EQTP
         citation = (
             "Peng L, Li L, Zeng X. "
@@ -49,7 +53,7 @@ class EQTP(EQTransformer):
         )
 
         # Initialize parent class
-        super().__init__(in_samples = in_samples, **kwargs)
+        super().__init__(in_samples=in_samples, **kwargs)
 
         # Override citation and labels for EQTP
         self._citation = citation
@@ -79,7 +83,6 @@ class EQTP(EQTransformer):
         self._rebuild_picking_branches(self.original_compatible, eps)
 
     def _rebuild_eqtp_components(self):
-
         self.encoder = Encoder(
             input_channels=self.in_channels,
             filters=self.filters,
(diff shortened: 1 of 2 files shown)


yetinam:
Hi @Pengling2022 ,

thanks for the contribution! In general, I'm happy to add the model and I think the PR is already in good shape. So far, I only had a short look at it, but there's a few things to change:

- As CEED is already completely integrated into SeisBench, I think there's no need to add the NCEDC subset you've implemented, so I'd suggest removing this class.
- From what I can see, the code for EQTP is in large parts a copy of the EQT code. To keep following the DRY principle, it would be better to subclass EQT and only overwrite the relevant parts.
- There should be unit test. There's no need for anything complex, but the tests should check that the annotate and classify functions for EQT work and produce output as expected. You can find examples for the existing models in the repo. There should also be tests for the new augmentations.

Once that is done, I'll look at the code in more detail. Two additional things. First, it would be good to include pretrained weights in the SeisBench repo. If you can send me the weights, I can upload them. Second, there's a parallel PR for PhaseNet+ #373 with a similar approach. We should make sure both use the same data types for the polarity annotations. But no need to worry about this too much now, I'll comment on it when I'll do a more thorough review of your code and then we can find the best way to implement this.

seisbench/generate/labeling.py
```diff
@@ -454,6 +454,132 @@ def __str__(self):
...
+        self.shape = shape
+        self._labelshape_fn_mapper = {
+            "gaussian": gaussian_pick,
+            "triangle": triangle_pick,
+            "box": box_pick,
+        }
+        kwargs["dim"] = kwargs.get("dim", 1)
+        super().__init__(label_type="multi_class", **kwargs)
+
+    def label(self, X, metadata):
```
yetinam: Given this function is in large part a duplicate of `ProbabilisticLabeller.label`, I think it would be good to unify the two. From what I can see, the only thing differing is what exactly is written into the y value. Maybe the inheritance could be something like:
`PickLabeller`
-> `StreamPickLabeller` (would host most of the function with an if-statement switching between the labels`
----> `ProbabilisticLabeller` & `PolarityPlusLabeller` (essentially just setting a configuration parameter to ensure the function in the super class does the right thing.

seisbench/models/eqtp.py
```diff
@@ -0,0 +1,306 @@
...
+    _annotate_args["polarity_threshold"] = ("Polarity threshold", 0.3)
+    _annotate_args["polarity_dict"] = ("The position of the polarity vector represents the polarity", {
+        2: "N",
+        0: "U",
+        1: "D"
+    })
+
+    def __init__(
+            self,
+            in_channels=3,
```
yetinam: Can you delete all constructor parameters that aren't explicitly used (like `in_channels` and `in_samples`), so that they are just passed through using kwargs?

seisbench/models/eqtp.py
```diff
@@ -0,0 +1,306 @@
...
+    It is an extended version of the EQTransformer model, which builds upon its
+    phase picking capabilities by adding P-wave polarity determination functionality.
+
+    This model is designed for processing three-component seismic waveform data,
+    and can simultaneously output picking results for phases such as P-waves/S-waves
+    and polarity determination results for P-waves (Up U/Down D/Unknown N).
+
+    Implementation is adapted from the EQTransformer with SeisBench GitHub repository (https://github.com/seisbench/seisbench).
+
+    the EQTP model can be instantiated via the `from_pretrained("NCEDC")` method.
```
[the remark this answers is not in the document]
yetinam: Fixed it.

yetinam:
Hi @Pengling2022 ,

sorry for the delay, but I've finally gotten around to give this PR a detailed look. Overall, it looks good to me and I'm happy to add this to SeisBench. I've added a couple of small comments that I'd ask you to address, but once that's done we're ready to merge.

The CI failure is related to formatting. You can fix this by installing the precommit hooks as detailed in the CONTRIBUTING.MD. However, if you should have trouble with that it's no problem for me to quickly fix the formatting questions myself.

PR #387, opened 2025-12-08
Refactor Annotation Pipelines

miili:
This PR replaces the `asyncio.Queue` with a `AsyncGenerator` pipeline - simplifying the code.

Furthermore, `NamedTuples` are used to pass data through the pipeline, replacing obscure tuples.

### DIFF
--- a/seisbench/models/base.py
+++ b/seisbench/models/base.py
@@ -1829,9 +1829,7 @@ class WaveformModel(SeisBenchModel, ABC):
             fragments = seisbench.util.pad_packed_sequence(data)
         else:
             fragments = np.array(data)
-        fragments = torch.tensor(fragments, device=self.device, dtype=torch.float32)
-        # fragments = torch.as_tensor(fragments, dtype=torch.float32, device=self.device)
-        # fragments = torch.from_numpy(fragments).float().to(self.device)
+        fragments = torch.as_tensor(fragments, dtype=torch.float32, device=self.device)
 
         with torch.no_grad():
             preprocessed = self.annotate_batch_pre(fragments, argdict=argdict)


seisbench/models/base.py
```diff
@@ -1519,273 +1478,363 @@ def _cut_fragments_point(self, t0, block, stations, argdict):
...
+        if len(offsets) == 0:
             seisbench.logger.warning(
                 "Parts of the input stream consist of fragments shorter than the number "
                 "of input samples. Output might be empty."
             )
-            return
+            return []
 
-        bucket_id = np.random.randint(1000000)
+        bucket_id = np.random.randint(1e9)
```
yetinam: This `bucket_id` is never used, is it? (It also sounds like a terrible idea on my side if we ever used random numbers as hopefully unique identifiers.) If it's unused, we could remove it from the code.

seisbench/models/base.py
```diff
@@ -1519,273 +1478,363 @@ def _cut_fragments_point(self, t0, block, stations, argdict):
...
+        argdict: dict,
+    ) -> PredictionsStacked:
         """
         Reassembles array predictions into numpy arrays.
         """
-        window, metadata = elem
-        key, t0, s, len_starts, stations, bucket_id, in_samples, pred_sample = metadata
-        buffer[key].append(elem)
+        n_predictions = len(predictions)
+        if n_predictions == 0:
```
yetinam: Why would that ever happen?

seisbench/models/base.py
```diff
@@ -1519,273 +1478,363 @@ def _cut_fragments_point(self, t0, block, stations, argdict):
...
         """
+        data = [window.data for window in segments]
         if self.allow_padding:
-            fragments = seisbench.util.pad_packed_sequence(buffer)
+            fragments = seisbench.util.pad_packed_sequence(data)
         else:
-            fragments = np.stack(buffer, axis=0)
-        fragments = torch.tensor(fragments, device=self.device, dtype=torch.float32)
+            fragments = np.array(data)
+        fragments = torch.from_numpy(fragments).to(self.device).float()
```
yetinam: I'd use `torch.to_tensor(fragments, device=self.device, dtype=torch.float32)`. This will automatically minimize the number of copies.

seisbench/models/base.py
```diff
@@ -1960,15 +2013,27 @@ def _trim_nan(x):
...
+        nan_mask = ~np.isnan(x)
+        valid = np.nonzero(nan_mask)[0]
+        nan_mask[valid[0] : valid[-1]] = True
+        end = x.size
 
-        return x, valid[0], _end - (1 + valid[-1])
+        return x[nan_mask], valid[0], end - (1 + valid[-1])
 
-    def _recursive_torch_to_numpy(self, x):
+    @staticmethod
```
yetinam: This function seems unused. I think it's from the other PR? I'd suggest removing it here.

seisbench/models/base.py
miili: removed it.

seisbench/models/base.py
```diff
@@ -1519,273 +1478,363 @@ def _cut_fragments_point(self, t0, block, stations, argdict):
...
-        window, metadata = elem
-        t0, s, len_starts, stations, bucket_id = metadata
-        key = f"{t0}_{'__'.join(stations)}"
+        # Sort by start (overwrite keys to make sure window is never used as key)
+        segments = sorted(segments, key=lambda seg: seg.window_offset)
+        meta = segments[0]
+        data = np.array([seg.data for seg in segments])
+        if data.ndim == 1:
+            data = data.reshape(-1, 1)
+        pred_time = meta.start_time + self.pred_sample / argdict["sampling_rate"]
```
yetinam: This won't fail for any currently implemented model, but in principle, this should be `meta.pred_sample`. At this point, you'll get an error though from the changes in `_get_in_pred_samples` (see below).

seisbench/models/base.py
```diff
@@ -1519,273 +1478,363 @@ def _cut_fragments_point(self, t0, block, stations, argdict):
...
-
                 preds = self.annotate_batch_post(
-                    preds, piggyback=piggyback, argdict=argdict
+                    preds,
+                    piggyback=piggyback,
+                    argdict=argdict,
                 )
         finally:
-            if train_mode:
+            if train_mode and not self.training:
```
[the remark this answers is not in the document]
miili: simplified and moved up to `annotate_async`

seisbench/models/base.py
```diff
@@ -2713,3 +2781,9 @@ def forward(self, input, state=None):
 
         # Keep second argument for consistency with PyTorch LSTM
         return output, None
+
+
+def exception_hanlder(loop, context):
```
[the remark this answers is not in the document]
miili: corrected.

seisbench/models/base.py
miili: changed. please check

seisbench/models/base.py
```diff
@@ -1519,273 +1478,363 @@ def _cut_fragments_point(self, t0, block, stations, argdict):
...
+        out_queue: asyncio.Queue[list[TraceSegment] | None] = asyncio.Queue()
 
-            output = ((pred_rate, pred_time, preds), stations)
+        async def drain_buffer(force: bool = False) -> None:
+            if not force and any(len(v) >= batch_size for v in buffer.values()):
+                return
 
-            del buffer[key]
+            for size in list(buffer.keys()):
+                while len(buffer[size]):
```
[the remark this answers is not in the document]
miili: Good catch! Please review the fix carefully.

miili:
> The PR looks good to me now. Last thing is the `tensor`/`as_tensor`/`from_numpy` question above. Could you give me a heads-up on the rational and remove the two options that are currently commented out?

That was a leftover, sorry about that; `as_tensor` it is.

PR #411, opened 2026-03-24
Add DeepSubDAS model

yetinam:
Adds the DeepSubDAS model by Xiao et al. (2026). PR should only be merged once the model weights are available in the SeisBench repository.

As a supplementary PR, an example notebook for DAS models should be provided.


yetinam:
Weights are online now

PR #420, opened 2026-05-25
Add DKPN waveform model integration

mbagagli:
This PR adds `DKPN` as a first-class SeisBench waveform model.

DKPN, Domain-Knowledge PhaseNet, is a seismic phase picker that reuses the PhaseNet U-Net architecture with domain-knowledge preprocessing. It accepts raw 3C waveform streams through `annotate()`/`classify()` and converts them to the five DKPN feature channels: `CFZ`, `CFN`, `CFE`, incidence `I`, and modulus `M`.

Citation:

Lomax, A., Bagagli, M., Gaviano, S., Cianetti, S., Jozinović, D., Michelini, A., ... & Giunchi, C. (2024). Effects on a deep-learning, seismic arrival-time picker of domain-knowledge based preprocessing of input seismograms. *Seismica*, 3(1). doi:10.26443/seismica.v3i1.1164

## Changes

- Added `seisbench.models.DKPN`
- Implemented DKPN as a `PhaseNet` subclass to avoid duplicated architecture code
- Added shared DKPN feature extraction for stream inference and generator training
- Added `seisbench.generate.DKPNPreProcessor` for training pipelines
- Preserved DKPN-specific normalization in `DKPN.annotate_batch_pre()`
- Added annotation, classification, save/load, and metadata support
- Updated docs, model tables, tests, and DKPN training example notebook
- Added a dedicated test suite in tests/models/train_dkpn.py

**All tests and pre-commits checks passed!**

The DKPN code path is ready for pretrained weights, but this PR does not include checkpoint files used in the published paper.

Should checkpoint metadata/conversion material be included here, or should pretrained DKPN weights be handled separately through the SeisBench model repository?


seisbench/generate/augmentation.py
```diff
@@ -6,6 +6,173 @@
...
+
+    def __call__(self, state_dict):
+        x, metadata = state_dict[self.key[0]]
+        metadata = copy.deepcopy(metadata)
+
+        sampling_rate = self._sampling_rate(metadata)
+        component_order = self._component_order(x, metadata)
+        waveforms = self._ordered_waveforms(x, component_order)
+
+        from seisbench.models.dkpn import _DKPNFeatureExtractor
```
yetinam: Please move all imports to the top of the file.

yetinam: In general, SeisBench does not allow for imports between `generate` and `models` (see [here](https://github.com/seisbench/seisbench/blob/main/CONTRIBUTING.md#code-structure)). I'd be hesitant to break this concept, because it might lead to circular imports at some point. Maybe you could move the feature extractor to `util`, so you can import it from both `models` and `generate`? I think the same pattern appears again further down for component ordering.

seisbench/generate/augmentation.py
```diff
@@ -6,6 +6,173 @@
...
+                    f"component_order='{component_order}'."
+                )
+
+        return component_order
+
+    def _ordered_waveforms(self, x, component_order):
+        if x.shape[0] != len(component_order):
+            x = x.T
+
+        from seisbench.models.dkpn import DKPN
```
yetinam: See comments above

seisbench/models/dkpn.py
```diff
@@ -0,0 +1,482 @@
...
+    license by Matteo Bagagli, Anthony Lomax, Sonja Gaviano, and the
+    SOME project.
+
+    .. document_args:: seisbench.models DKPN
+    """
+
+    _feature_component_order = "ZNEIM"
+    _raw_component_order = "ZNE"
+
+    _dkpn_defaults = {
```
yetinam: This seems to be a duplicate to me, given these parameters should all be defined through the `_annotate_args`. Most of them are there already, but some are missing. If you add them to `_annotate_args`, you can remove `_dkpn_defaults`. Right now, they will overwrite the `_annotate_args`, rendering the documentation potentially incorrect.

seisbench/models/dkpn.py
```diff
@@ -0,0 +1,482 @@
...
+        "mode": "rms",
+        "clip": -999,
+        "log": True,
+        "normalize": False,
+        "polarization_win_len": 1,
+        "use_amax_only": False,
+        "blinding": (250, 250),
+    }
+
+    _annotate_args = PhaseNet._annotate_args.copy()
```
yetinam, line 62: These parameters can all be overwritten when calling `annotate`/`classify`. Is this intended for all of them? Otherwise, they should be moved to arguments of `__init__`.

seisbench/models/dkpn.py
```diff
@@ -0,0 +1,482 @@
...
+    ):
+        citation = (
+            "A. Lomax, M. Bagagli, S. Gaviano, S. Cianetti, D. Jozinović, "
+            "A. Michelini, C. Zerafa, C. Giunchi (2024). Effects on a "
+            "deep-learning, seismic arrival-time picker of domain-knowledge "
+            "based preprocessing of input seismograms. Seismica, 3(1). "
+            "doi:10.26443/seismica.v3i1.1164"
+        )
+
+        if default_args is None:
```
yetinam: This block should be moved as it interferes with the way `_annotate_args` work.

seisbench/models/dkpn.py
```diff
@@ -0,0 +1,482 @@
...
+        model_args["in_channels"] = self.in_channels
+        model_args["classes"] = self.classes
+        model_args["phases"] = self.labels
+        model_args["sampling_rate"] = self.sampling_rate
+
+        return model_args
+
+
+class _DKPNFeatureExtractor:
+    default_args = {
```
yetinam: It seems that the default args are defined in multiple places. Maybe you could instead move them to the top of the class in a constant like `DKPN_DEFAULT_ARGS`. Otherwise, future updates might make them inconsistent by changing them in some but not all places.

seisbench/models/dkpn.py
```diff
@@ -0,0 +1,482 @@
...
+        "use_amax_only": False,
+    }
+
+    def __init__(self, **kwargs):
+        values = {**self.default_args, **kwargs}
+        self.__dict__.update(values)
+        self.eps = 1e-10
+
+    def matrix_cfs(self, waveforms, sampling_rate):
+        waveforms = waveforms.astype(np.float64, copy=True)
```
yetinam: I think the copy operation is not required, because the line below triggers a copy in either case.

seisbench/models/dkpn.py
```diff
@@ -0,0 +1,482 @@
...
+        waveforms = waveforms - np.mean(waveforms, axis=1, keepdims=True)
+        waveforms = waveforms / (np.std(waveforms, axis=1, keepdims=True) + self.eps)
```
yetinam: Can be changed to an inplace operation with `/=` to avoid an extra copy

seisbench/models/dkpn.py
```diff
@@ -0,0 +1,482 @@
...
+
+        return incidence, modulus
+
+
+class _FBSummary:
+    def __init__(
+        self,
+        data,
+        npts=3001,
+        sampling_rate=100.0,
```
yetinam: Nit: It might be better not to define a default sampling rate so users are required to pass the correct one.

yetinam:
Hi @mbagagli ,

thanks a lot for the contribution! Overall, it's looking really good and I think it's a great solution to reuse the GroupingHelper to generate the input traces. Thanks also for adding the training notebooks and the detailed unit tests. I left a few comments on the code, mostly about a refactor to avoid imports across packages. Let me know if you have questions about that.

For the example notebook, can you add it in the README.MD, the examples README.MD, and the examples page of the documentation? Please let the colab link point at the main branch of the SeisBench repo. Even though that won't work before merging, it'll be long-term stable afterwards.

Regarding the weights, can you send them to me by mail, so I can upload them in the SeisBench repository?

seisbench/models/dkpn.py
mbagagli: Right, deleted as results of previous alignment with  `_default_args` behavior

seisbench/generate/augmentation.py
mbagagli: done

seisbench/generate/augmentation.py
mbagagli: done

mbagagli: done

seisbench/models/dkpn.py
mbagagli: I've left sampling_rate=100.0 as class default on initialization (as PhaseNet does), and left that value also on the pretrained weights, as otherwise it would break the DKPN behavior

seisbench/models/dkpn.py
mbagagli: done

seisbench/models/dkpn.py
mbagagli: done

seisbench/models/dkpn.py
mbagagli: see at the bottom for discussion

seisbench/models/dkpn.py
mbagagli: see at the bottom for discussion

seisbench/models/dkpn.py
mbagagli: see at the bottom for discussion

yetinam: Hi @mbagagli , did you post the discussion at the bottom. It seems like I can't find it.

mbagagli:
Ciao Jannes! Thanks for the revision!
I've pushed everything just now, I needed to finish few things beforehand.
I've tackled your comments and replied briefly on each point (also for my later reference), but with the main discussion here in this comment.

I split the DKPN defaults into two layers to avoid duplicated or misleading `default_args`. `_DKPN_FEATURE_DEFAULTS` is now the single internal source for scientific feature-generation settings such as `fp_stabilization`, `freqmin`, and `t_long`; these are model configuration and are stored through `model_args`, so they round-trip with saved/pretrained metadata. `_annotate_args` only exposes runtime annotation/classification controls, via `_DKPN_ANNOTATE_DEFAULTS`, such as thresholds, overlap, and blinding. 

This should keeps per-call `annotate()`/`classify()` kwargs from accidentally changing the preprocessing definition of a pretrained model. `DKPNPreProcessor` imports the same internal DKPN defaults/extractor so training and stream inference share one numerical feature path; if preferred, we can move those internals to a small shared private utility module, but the current version keeps the implementation compact and DRY inside the package.

I'm sending you soon  the weights and if any more comments on the new commits are needed, please let me know.

Cheers

seisbench/models/dkpn.py
mbagagli: Hi @yetinam , sorry it needed one additional tweak and testing before final pushes, I've commented now at the bottom.

yetinam:
I just incorporated some final changes and uploaded the weights to the SeisBench repository. Given the last change was only to the docstring, I'll merge without another CI pass.

Thanks a lot for this contribution! It'll be part of SeisBench v0.12 which I'll release later today.

PR #435, opened 2026-08-23
Add optional progress bar to annotate

### LINKED ISSUE #211: Progress bar parameter for annotate
Running these models can take a long time, so a little progress bar for annotate would be a nice thing to have. I'd like to try adding it. What do you think? I'd add a parameter to annotate to set the progress bar off by default.

manduinca:
Closes #211

You warned in the issue that the tricky part is knowing the output size upfront. The bar sidesteps that by counting groups of traces instead of windows: `group_stream` already returns them as a list before annotation starts, so `tqdm` gets a real total.

It advances in the loop that consumes the finished streams, not the one that queues work. That matters — putting it in `_iter_groups` would have filled the bar while the model was still predicting, since that stage only does the cheap preprocessing.

Uses `tqdm` as you asked, following the `pbar` pattern already in `das_base.py`.

I defaulted it to `False` so nothing changes for existing callers. Say the word and I'll flip it to `True` for consistency with the DAS path.

The test replaces `tqdm` with a double to check the total matches the group count and that no bar is built unless asked.


yetinam:
Hi @manduinca ,
thanks for the comprehensive PR (and sorry for the long wait). I just merged it. It's indeed a very elegant solution. Naturally, if the different groups have different sizes, this can still lead to unsteady progress bars, but it gives a good first orientation. For now, let's keep the default value as `False`. I think in contrast to DAS, most applications of classical models are fast enough that a progress bar isn't required by default.

